PR #14, opened 2020-09-05
add example

### LINKED ISSUE #13: Add simple binary example
We should have an example in this repo similar to `deno_lint`'s example binary called [`dlint`](https://github.com/denoland/deno_lint/tree/master/examples/dlint).

zhmushan:
close #13

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,12 @@ description = "doc generation for deno"
 authors = ["the Deno authors"]
 license = "MIT"
 
+[lib]
+name = "deno_doc"
+
+[[example]]
+name = "ddoc"
+
 [dependencies]
 futures = "0.3.5"
 lazy_static = "1.4.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,5 +24,6 @@ termcolor = "1.1.0"
 regex = "1.3.9"
 
 [dev-dependencies]
+clap = "2.33.3"
 tokio = { version = "0.2.22", features = ["full"] }
 url = "2.1.1"
--- a/examples/ddoc/main.rs
+++ b/examples/ddoc/main.rs
@@ -1,0 +1,89 @@
+use deno_doc::parser::DocFileLoader;
+use deno_doc::{DocError, DocNodeKind, DocParser, DocPrinter};
+use futures::executor::block_on;
+use futures::FutureExt;
+use std::env::{args, current_dir};
+use std::fs::read_to_string;
+use swc_ecmascript::parser::{Syntax, TsConfig};
+use tokio::macros::support::{Future, Pin};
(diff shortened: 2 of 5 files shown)


examples/ddoc/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    let fetcher = self.clone();
+    let module_url = Url::parse(specifier).unwrap();
+    async move {
+      let url_scheme = module_url.scheme();
+      let is_local_file = url_scheme == "file";
+
+      if is_local_file {
+        Ok(fetcher.fetch_local_file(&module_url))
+      } else {
+        Ok(fetcher.fetch_remote_source(&module_url))
```
bartlomieju, line 57: Let's not add handling for remote sources at all - please return error when user tries to load remote source

examples/ddoc/main.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl SourceFileFetcher {
+  pub fn new() -> Box<Self> {
+    Box::new(Self {})
+  }
+
+  fn fetch_local_file(&self, module_url: &Url) -> String {
+    let path = module_url.to_file_path().unwrap();
+    read_to_string(path).unwrap()
+  }
```
bartlomieju, line 22: This impl block is superfluous

examples/ddoc/main.rs
```diff
@@ -0,0 +1,89 @@
...
+        Ok(fetcher.fetch_remote_source(&module_url))
+      }
+    }
+    .boxed_local()
+  }
+}
+
+fn main() {
+  let args: Vec<String> = args().collect();
+  let target = args.get(1).unwrap();
```
bartlomieju, line 66: Maybe add `clap` for nicer CLI?

bartlomieju (changes requested):

zhmushan:
@bartlomieju done

Cargo.toml
```diff
@@ -18,5 +24,6 @@ termcolor = "1.1.0"
 regex = "1.3.9"
 
 [dev-dependencies]
+clap = "2.33.3"
```
ry, line 27: It’s just another thing to keep in-sync. I’d think it’d be better if the example was just a minimal example of integrating with deno_doc

lucacasonato: This is a dev dependency. It doesn't need to be kept in sync. Just like for `deno_lint`.

zhmushan: This is my first time using Clap, and it is indeed a good start for creating cli applications.  But I have no idea whether to recommend it to others as an example. This is a trivial matter and users will not be confused by it.

bartlomieju (approved):
LGTM, thanks @zhmushan

PR #18, opened 2020-09-07
public parse_module & ModuleDoc

### LINKED ISSUE #17: Provide a way to parse modules in the DocParser
In some cases, I want to handle `reexports` manually. Could we provide a method which returns `ModuleDoc`?

zhmushan:
If we public `parse_module`, the related struct should also be public

close #17


bartlomieju (approved):
LGTM

PR #19, opened 2020-09-07
fix: Export as another name

zhmushan:


bartlomieju (approved):
LGTM, thanks @zhmushan

PR #22, opened 2020-09-09
fix: export interface and class with same name

### LINKED ISSUE #15: Exported interface and class with same name only interface is documented
[in this example](https://doc.deno.land/https/raw.githubusercontent.com/denosaurs/denord/26ae2cc4c0f8c03b14d2b36eaa74b9e6d49665ed/mod.ts#ShardManager) documentation only shows the interface for `ShardManager`, but actually what is exported from the file is a `ShardManager` interface & class.

zhmushan:
close #15


lucacasonato (approved):
LGTM - great that this fix is so small. Thanks

PR #29, opened 2020-09-25
refactor: add Deserialize derive to nodes

zhmushan:
When using `deno_doc` as a crate, it is useful to support deserialize, such as deserializing the previously generated JSON data and formatting it through `DocPrinter`


bartlomieju (approved):
LGTM, thanks @zhmushan

PR #38, opened 2020-10-11
chore: make colors public

zhmushan:
Finally we will move `printer` back into `deno/cli`, so it is necessary to be able to enable & disable color externally.

(It is annoying to format the interface property alone and cannot carry colors

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,7 +7,7 @@ extern crate lazy_static;
 extern crate serde_json;
 
 pub mod class;
-mod colors;
+pub mod colors;
 mod display;
 pub mod r#enum;
 pub mod function;


bartlomieju:
@zhmushan thanks for the PR, but I'm not sure if making `colors` public now is a good idea. I'd rather do it in one go, when moving printing logic to CLI (if ever, I actually think it's better to keep it here).

zhmushan:
@bartlomieju if I just want to format for example the interface properties, and make it have colors, is there any way?

bartlomieju:
@zhmushan [`DocPrinter`](https://docs.rs/deno_doc/0.1.11/deno_doc/printer/struct.DocPrinter.html) takes a parameter to decide if it should colorize or not

zhmushan:
@bartlomieju Unfortunately, DocPrinter can only format an entire DocNode object, not just a part of it

ry:
@zhmushan - I'm closing this because it's old and stale. If you still want this colors module made public, please open a new PR.

PR #49, opened 2020-11-14
chore: remove serde_derive from dependencies

magurotuna:
`serde_derive` is listed in dependencies, but it is not used at all. Instead we are using `serde::{Serialize, Deserialize}` from `serde` (not `serde_derive`).
So in this PR, I removed `serde_derive` from dependencies.


bartlomieju (approved):
LGTM, thanks

PR #56, opened 2020-11-22
feat: allow deep reexports

### LINKED ISSUE #54: Doesn't support deep reexports
### Example:
```typescript
// foo.ts
export const foo = "foo";
```
```typescript
// bar.ts
export * from "./foo.ts";
```
```typescript
// baz.ts
export * from "./bar.ts";
```
`ddoc baz.ts` should output `const foo: string`, but outputs nothing.

This is already mentioned in #4, but I thought I should create a separate issue before I try and work on it.

Liamolucko:
fixes #54

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -18,6 +18,7 @@ use crate::DocNodeKind;
 use crate::ImportDef;
 use crate::Location;
 use futures::Future;
+use futures::FutureExt;
 use regex::Regex;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, although I'm a little hesitant on landing this because of unsure future of `DocFileLoader` trait. We should update doc website to reflect this change after landing.

CC @lucacasonato

PR #60, opened 2020-11-22
fix: only include declarations in docs if the file is ambient

### LINKED ISSUE #57: Renamed export retains old name
### Example:
```typescript
declare function foo(): void;
export { foo as bar };
```
#### Expected:
`function bar(): void`
#### Actual:
`function foo(): void`

This is only the case with declarations. If `foo()` was actually implemented here, the output would be correct.

Liamolucko:
Previously, the `declare` keyword was taken to indicate a global variable and shown in the documentation. However, that is only be the case for ambient declarations - in modules, this just indicates a private value.

Partially fixes #57

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -595,14 +595,19 @@ impl DocParser {
   ) -> Vec<DocNode> {
     let mut unexported_doc_map: HashMap<String, DocNode> = HashMap::new();
     let mut doc_entries: Vec<DocNode> = vec![];
+    let mut ambient_entries: Vec<DocNode> = vec![];
+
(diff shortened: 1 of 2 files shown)


Liamolucko:
This actually includes all the changes from #59, which I guess it probably shouldn't, but the goal was to fix #57 and those changes are needed to fix it. Maybe I should undo that?

Liamolucko:
I think I'll leave it to avoid merge conflicts.

bartlomieju:
@Liamolucko please rebase

bartlomieju (approved):
Cool, thanks @Liamolucko

PR #63, opened 2020-11-23
fix: support default exports for statements declared earlier

### LINKED ISSUE #62: Doesn't detect reexported default exports
### Example
```typescript
function foo() {}
export default foo;
```
`deno doc` should show `foo` as the default export, but shows nothing.

Liamolucko:
fixes #62

Note: this includes the changes from #59, since this kind of needs those changes (at least not to have merge conflicts).


Liamolucko:
This is basically an extension of #59, actually - I think I'll combine these PRs, particularly with the weirdness around including its changes here.

PR #67, opened 2020-12-09
chore: update deps

lucacasonato:


bartlomieju (commented):
Version should be bumped as well

bartlomieju (approved):

PR #72, opened 2020-12-27
feat: Add support for filtering class and interface fields

uki00a:
This PR implements `detail output for class fields` listed at the issue #4:

**detail output for interface fields:**

```shell
$ ./target/debug/examples/ddoc ../deno/std/encoding/csv.ts ParseOptions.columns
Defined in file:///home/uki00a/ghq/github.com/uki00a/deno/std/encoding/csv.ts:363:2

const columns: string[] | ColumnOptions[]
  If you provide `string[]` or `ColumnOptions[]`, those names will be used for header definition.
```

**detail output for class fields:**

```shell
$ ./target/debug/examples/ddoc ../deno/std/io/bufio.ts BufReader.read
Defined in file:///home/uki00a/ghq/github.com/uki00a/deno/std/io/bufio.ts:124:2

async function read(p: Uint8Array): Promise<number | null>
  reads data into p.
  It returns the number of bytes read into p.
  The bytes are taken from at most one Read on the underlying Reader,
  hence n may be less than len(p).
  To read exactly len(p) bytes, use io.ReadFull(b, p).
```

### DIFF
--- a/src/class.rs
+++ b/src/class.rs
@@ -21,6 +21,8 @@ use crate::ts_type::{
 };
 use crate::ts_type_param::maybe_type_param_decl_to_type_param_defs;
 use crate::ts_type_param::TsTypeParamDef;
+use crate::variable::VariableDef;
+use crate::DocNode;
 use crate::Location;
 use crate::ParamDef;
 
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM! Nice tests @uki00a

PR #73, opened 2020-12-27
chore: Add .editorconfig

uki00a:
This PR adds `.editorconfig` based on [the one in denoland/deno](https://github.com/denoland/deno/blob/master/.editorconfig).


lucacasonato (approved):
Thanks

PR #85, opened 2021-03-01
fix: Don't use the same syntax for all files

Liamolucko:
Ref denoland/deno#9617

BREAKING CHANGE: Changes `DocFileLoader`, `parse`, and `parse_with_reexports` APIs

### DIFF
--- a/examples/ddoc/main.rs
+++ b/examples/ddoc/main.rs
@@ -30,7 +30,7 @@ impl DocFileLoader for SourceFileFetcher {
   fn load_source_code(
     &self,
     specifier: &str,
-  ) -> Pin<Box<dyn Future<Output = Result<String, DocError>>>> {
+  ) -> Pin<Box<dyn Future<Output = Result<(Syntax, String), DocError>>>> {
     let module_url = Url::parse(specifier).unwrap();
     async move {
       let url_scheme = module_url.scheme();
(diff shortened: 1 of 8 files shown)


examples/ddoc/main.rs
```diff
@@ -30,15 +30,18 @@ impl DocFileLoader for SourceFileFetcher {
   fn load_source_code(
     &self,
     specifier: &str,
-  ) -> Pin<Box<dyn Future<Output = Result<String, DocError>>>> {
+  ) -> LocalBoxFuture<Result<(Syntax, String), DocError>> {
```
bartlomieju, line 33: I'm curious if `LocalBoxFuture` is required here... In Deno most of the futures are not `Send` and we do not use this `LocalBoxFuture`. Doesn't `Pin<Box<...>>` work? You can use `.boxed_local()` on the future to turn make type checker happy.

Liamolucko: `LocalBoxFuture` is just a type alias for `Pin<Box<dyn Future>>` - this was to satisfy a nightly clippy lint about types being too long.

bartlomieju: Fair enough; I'm gonna land this PR and cut a release of `deno_doc` soon.

Issue #102, opened 2021-05-15
User-defined type guards aren't supported


metanomial:
Deno doc renders the predicate of user-defined type guards (`function (X): X is Y`) as `<UNIMPLEMENTED>` in plaintext output, or `_notimpl_` in JSON output.

Example:
https://doc.deno.land/https/gist.githubusercontent.com%2Fmetanomial%2F9224b007807df8bdb9c32e663218f75c%2Fraw%2Fec7caf16ad0e087dcc097184c5a131061d510787%2Fdeno-doc-type-predicates.ts

PR #105, opened 2021-06-12
feat: handle user-defined type guards

magurotuna:
This PR deals with TypeScript user-defined type predicate and assertion, which were unimplemented as reported in #102. Additionally, added `pretty_assertions` crate as a dev-dependency so we can easily spot the difference between expected and actual JSON output.
As far as I understand, to address #102 completely, we need to fix [doc_website](https://github.com/denoland/doc_website) as well. I will work on it later.

#### Demo

```ts
export function f1(val1: A | B): val1 is A {}
export function f2(val2: any): asserts val2 is string {}
export function f3(val3: any): asserts val3 {}
export function assertIsDefined<T>(val4: T): asserts val4 is NonNullable<T> {
  if (val === undefined || val === null) {
    throw new AssertionError(
      `Expected 'val' to be defined, but received ${val}`
    );
  }
}
export class C {
  isSomething(): this is Something {
    return this instanceof Something;
  }
}
```

Running `cargo run --example ddoc foo.ts`, we get:

```
Defined in file:///home/yusuktan/Repos/github.com/magurotuna/deno_doc/foo.ts:4:0

function assertIsDefined<T>(val4: T): asserts val4 is NonNullable<T>

Defined in file:///home/yusuktan/Repos/github.com/magurotuna/deno_doc/foo.ts:1:0

function f1(val1: A | B): val1 is A

Defined in file:///home/yusuktan/Repos/github.com/magurotuna/deno_doc/foo.ts:2:0

function f2(val2: any): asserts val2 is string

Defined in file:///home/yusuktan/Repos/github.com/magurotuna/deno_doc/foo.ts:3:0

function f3(val3: any): asserts val3

Defined in file:///home/yusuktan/Repos/github.com/magurotuna/deno_doc/foo.ts:11:0

class C

  isSomething(): this is Something
```

### DIFF
--- a/src/ts_type.rs
+++ b/src/ts_type.rs
@@ -830,10 +830,10 @@ pub struct TsTypeDef {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-#[serde(untagged, rename_all = "camelCase")]
+#[serde(tag = "type", rename_all = "camelCase")]
 pub enum ThisOrIdent {
   This,
-  Ident(String),
+  Identifier { name: String },
 }
 
 impl From<&TsThisTypeOrIdent> for ThisOrIdent {
@@ -841,7 +841,9 @@ impl From<&TsThisTypeOrIdent> for ThisOrIdent {
     use TsThisTypeOrIdent::*;
     match other {
       TsThisType(_) => Self::This,
-      Ident(ident) => Self::Ident(ident.sym.to_string()),
+      Ident(ident) => Self::Identifier {
+        name: ident.sym.to_string(),
+      },
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/ts_type.rs
```diff
@@ -803,6 +824,64 @@ pub struct TsTypeDef {
...
+pub struct TsTypePredicateDef {
+  /// (1) Whether the predicate includes `asserts` keyword or not
+  pub asserts: bool,
+
+  /// (2) The term of predicate
+  pub param: ThisOrIdent,
+
+  /// (3) The type against which the parameter is checked
+  pub r#type: Option<Box<TsTypeDef>>,
+}
```
bartlomieju, line 867: Beautiful 👍

bartlomieju (approved):
LGTM, no comments - cleanly implemented with great tests exercising changes. 

Let's make sure to update the website for this feature before 1.12 comes out

PR #106, opened 2021-06-14
chore: upgrade swc_ecmascript

magurotuna:
To fix https://github.com/denoland/deno/issues/10954 and https://github.com/denoland/deno/issues/10893 we need to upgrade swc, so this commit does it so that all the related crates will have the same version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,6 +29,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


magurotuna:
CC @bartlomieju

bartlomieju (approved):
LGTM, thank you @magurotuna!

PR #124, opened 2021-08-11
refactor: migrate to deno_graph and other wasm prep-work

kitsonk:
This integrates `deno_graph` and lays down the prep work for exposing a web assembly API for `deno_graph`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,6 +36,10 @@ pub use parser::DocError;
 pub use parser::DocParser;
 pub use printer::DocPrinter;
 
+#[cfg(feature = "wee_alloc")]
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
 #[cfg(test)]
 mod tests;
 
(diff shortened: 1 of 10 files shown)


kitsonk:
This will be a breaking change with its integration into CLI, and I will do a refactor there to get it re-integrated again.

bartlomieju (approved):

src/lib.rs
```diff
@@ -36,6 +36,10 @@ pub use parser::DocError;
 pub use parser::DocParser;
 pub use printer::DocPrinter;
 
+#[cfg(feature = "wee_alloc")]
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```
bartlomieju, line 41: @lucacasonato claims that it's not really necessary - there's not much of a difference between default and `WeeAlloc`. I don't have a strong opinion either way.

kitsonk: I will remove it for now in a follow-up PR.  It saves about ~8k, but the wasm is a lot bigger for other reasons that 8k is a drop in the bucket and it can impact performance.

PR #125, opened 2021-08-13
feat: add TS/WASM API for crate

kitsonk:


.github/workflows/ci.yml
```diff
@@ -40,23 +40,34 @@ jobs:
           key: ${{ runner.os }}-${{ github.sha }}
           restore-keys: ${{ runner.os }}-
 
+      - name: ➡️  install Deno
+        uses: denoland/setup-deno@v1.0.0
```
lucacasonato, line 44: Lock to @v1 instead.

lucacasonato (approved):

lib/deno_doc.js
```diff
@@ -0,0 +1,419 @@
...
+      }
+    }
+  };
+  real.original = state;
+  CLOSURE_DTORS.register(real, state, state);
+  return real;
+}
+function __wbg_adapter_18(arg0, arg1, arg2) {
+  wasm
+    ._dyn_core__ops__function__FnMut__A____Output___R_as_wasm_bindgen__closure__WasmClosure___describe__invoke__h3c27f9a52bc1c519(
```
bartlomieju, line 209: Is this function name stable across compilations?

bartlomieju: Nvm, this file is auto generated...

mod.ts
```diff
@@ -0,0 +1,78 @@
...
+  load?(
+    specifier: string,
+    isDynamic: boolean,
+  ): Promise<LoadResponse | undefined>;
+  /** An optional callback that allows the default resolution logic of the
+   * module graph to be "overridden". This is intended to allow items like an
+   * import map to be used with the module graph. The callback takes the string
+   * of the module specifier from the referrer and the string URL of the
+   * referrer. The callback then returns a resolved URL string specifier. */
+  resolve?(specifier: string, referrer: string): string;
```
bartlomieju, line 49: Question here: in internal Deno infra this function has also `is_main` param which tells is the resolved file is the entry point - this is useful/required when used in combination with import map (which shouldn't be applied to the entry point, but should for all other files). Should we support it here?

src/lib.rs
```diff
@@ -1,48 +1,58 @@
...
-#[cfg(feature = "wee_alloc")]
-#[global_allocator]
-static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+cfg_if! {
+  if #[cfg(feature = "rust")] {
+    pub use node::DocNodeKind;
+    pub use parser::DocError;
+    pub use printer::DocPrinter;
+  }
+}
```
bartlomieju, line 43: This macro is useful

bartlomieju (approved):
LGTM, I have only minor questions, awesome work!

lib/deno_doc.js
kitsonk: It is "stable-ish"... as stable as we can get, and auto-generated as you noted.

mod.ts
kitsonk: I think it would be more a question for `deno_graph`, but, I would say that is an externality that shouldn't be a part of the API.  If there is additional contextual information, like what the root module is, that should be applied, that should be communicated to the resolver some other way.  Your specific scenario would be that the root specifier provided to create the module graph would need to be also passed to the import map.

PR #127, opened 2021-08-18
feat: export all public JS types

kitsonk:
Working with the types, I realised that we need all the types exported, and so instead of re-exporting some from the mod.ts, we should just export them and have them in `lib/types.d.ts` for importing as needed.


satyarohith (approved):
LGTM.

1057224030 (approved):

Issue #141, opened 2021-08-25
Structure JSDoc better


kitsonk:
Ref #111

I think we should introduce the following structure for `js_doc` which is current `Option<String>` in locations that support it:

```rust
/// This is incomplete, just illustrative...
pub enum JsDocTag {
  Param { name: String, type_ref: Option<String>, doc: Option<String> }
  Return { type_ref: Option<String>, doc: Option<String> }
  TypeRef { type_ref: Option<String>, doc: Option<String> }
  Deprecated { doc: Option<String> }
}

pub struct JsDoc {
  doc: Option<String>,
  tags: Vec<JsDocTag>,
}
```

PR #144, opened 2021-09-01
feat: structure JSDoc better

kitsonk:
Closes: #141


lucacasonato (commented):
Looks good overall.

I do wonder if we could not use the `jsdoc` crate for parsing? It is maintained by Kang. He is likely using it for STC, so I have relatively high confidence in it. Feels a lot more robust than regex matching.

kitsonk:
Ooops... ok, I will take a look at that. You question my Regex authoring? 😉

lucacasonato:
I would never dare question your regexp authoring 😅

kitsonk:
@lucacasonato I looked into the `jsdoc` crate... I think there are a few issues that mean integrating it right now would be problematic:

- For some reason any leading docs don't get stripped of their ` * `.  So this:

   ```js
    /**
     * Some leading comment
     * across multiple lines
     * 
     * @param {string} b comment
    ```

    Gets parsed out as `Some leading comment\n * across multiple lines\n *\n * `.  That could be intentional or unintentional, it is really hard to tell.
- Processing of the tags to put into structures is rather complicated, and it seems the tags only include what is meaningful for inferring types, versus generating documentation, like a lot of the tags can have comments/doc associated with them, but are just not part of the parse.  Getting stuff back out of the JSDoc "AST" into structures useful for documentation is complicated.
- The crate is changing a lot, I think it is far from a stable API at the moment, and upstreaming changes might is hard due to the lack of any sort of documentation in the code.

So, while it might make sense in the future to integrate `jsdoc`, it feels like it needs to be a bit more settled and a bit more mature.  Integrating it and switching over would be trivial in the future I suspect, but my opinion is that it isn't something we should do right now, and focus on stabilising the output which would make it even easier to modify how it is "parsed" in the future.

lucacasonato (approved):
Let's do this then 👍

irustm:
Guys, it would be cool to mention this PR in Breacking Changes in the release notes

PR #146, opened 2021-09-02
fix: prefer types when docing

kitsonk:
When doc'ing re-exports of JavaScript modules, prefer the type definitions, versus the JavaScript.

Requires https://github.com/denoland/deno_graph/pull/21 and will re-base when that is merged and crate published.


bartlomieju (approved):
LGTM, perfect, seems like a missing piece for https://github.com/denoland/deno/pull/11894

Issue #135, opened 2021-08-24
Support template literal types


kitsonk:
The following:

```ts
export type World = "world";
 
export type Greeting = `hello ${World}`;
```

`Greeting` ends up being just `"hello "` and is missing the template literal type.

kitsonk:
Upon looking at the code, the code was not properly handling template literal types, versus not supporting them, so this is actually a bug fix now.

PR #153, opened 2021-09-06
fix: properly handle template literal types

kitsonk:
Fixes: #135

This changes the format of the `"kind": "literal"` serialization of `tsType`, as the structure was not able to properly represent template literal types, so downstream consumers of the `DocNode` structure or serialization.

The `DocPrinter` though prints out the changes "properly".


src/ts_type.rs
```diff
@@ -1,52 +1,27 @@
...
-  TsThisTypeOrIdent, TsTupleType, TsType, TsTypeAnn, TsTypeLit, TsTypeOperator,
-  TsTypeParamInstantiation, TsTypePredicate, TsTypeQuery, TsTypeRef,
-  TsUnionOrIntersectionType, VarDeclarator,
-};
-
-// pub enum TsType {
-//  *      TsKeywordType(TsKeywordType),
-//  *      TsThisType(TsThisType),
-//  *      TsFnOrConstructorType(TsFnOrConstructorType),
-//  *      TsTypeRef(TsTypeRef),
```
bartlomieju, line 25: Thanks for cleaning this up

lib/types.d.ts
```diff
@@ -295,14 +301,19 @@ export interface LiteralDefNumber extends LiteralDefBase {
 
 export interface LiteralDefBigInt extends LiteralDefBase {
   kind: "bigInt";
-  number: bigint;
+  string: string;
```
bartlomieju, line 304: Is this change on purpose?

bartlomieju (approved):
LGTM, it's a bit unfortunate that it's a breaking change

lib/types.d.ts
kitsonk: Yeah, it was a mistake on my part, and I didn't notice it until I was making the other changes.  Silly me forgot there is no way to serialize a bigint to JSON as a bigint.

kitsonk:
It is a breaking change, because the original implementation was sadly incorrect on several levels.

Issue #133, opened 2021-08-23
Json information should have whether the property is decorated and what the decorator is.


andreespirela:
Json information should have whether the property is decorated and decorator properties such as name of the decorator

PR #156, opened 2021-09-07
feat: doc structures handle decorators

kitsonk:
Closes: #133

This also changes "<UNIMPLEMENTED>" to "[UNSUPPORTED]" across the board, as especially when rendering SSR, the unencoded `<` and `>` cause problems in display, and if the consumer doesn't sanitise it, you get strange results.


kitsonk:
Also, side note, I don't think swc supports [parameter decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators) and so the seralized structures don't currently support it.  This PR supports class, method/function, and property decorators.

andreespirela:
Is there a version set for this? Is there such thing as canary for deno doc? Thanks!

kitsonk:
For it as part of the CLI, the PR needs to be merged and a version of the `deno_doc` crate needs to be published, and the CLI needs to update to that crate version.  That will all likely take place before 1.14.  To use the JavaScript/WASM interface, you could do that right now by:

```ts
import { doc } from "https://raw.githubusercontent.com/kitsonk/deno_doc/kitsonk/issue133/mod.ts";

console.log(await doc("https://example.com/mod.ts"));
```

Though the branch will get deleted once the PR is merged.

andreespirela:
@kitsonk can you generate JSON with the wasm interface? Any guidence towards that?

kitsonk:
@andreespirela https://github.com/denoland/deno_doc#doc

andreespirela:
@kitsonk the API doesn’t accept import maps right? If not, do they take the main app import map into consideration?

kitsonk:
@andreespirela it does, in the sense of a resolver, but there isn't a default implementation of a resolver that takes an import map at the moment.

src/class.rs
```diff
@@ -64,6 +66,8 @@ pub struct ClassPropertyDef {
   pub ts_type: Option<TsTypeDef>,
   pub readonly: bool,
   pub accessibility: Option<deno_ast::swc::ast::Accessibility>,
+  #[serde(skip_serializing_if = "Vec::is_empty")]
+  pub decorators: Vec<DecoratorDef>,
```
bartlomieju, line 70: Does it make sense to skip serializing if vector is empty? We don't do that for any other elements

src/decorators.rs
```diff
@@ -0,0 +1,93 @@
...
+use serde::Serialize;
+use std::fmt::Display;
+use std::fmt::Formatter;
+use std::fmt::Result as FmtResult;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct DecoratorDef {
+  pub name: String,
+  #[serde(skip_serializing_if = "Vec::is_empty")]
```
bartlomieju, line 21: Ditto

bartlomieju (approved):
LGTM, but I'm unsure if we should skip serializing the field if there's no decorator

src/class.rs
kitsonk: I want to migrate to it in the future, and since this is additive it isn't a breaking change.  Having loads of properties on objects with empty arrays has a non-negligible impact on performance, when you start doing that by thousands ands thousands of nodes that can be in an JSON structure.  Any empty array adds no information and just causes "bloat".  We don't print out `there are no methods` in we format it for display, so it doesn't increase readability or usability.

bartlomieju: Okay, that's fair.

Issue #134, opened 2021-08-24
Support mapped types


kitsonk:
The following:

```ts
export type Partial<T> = {
  [P in keyof T]?: T[P];
};
```

Results in the type side being set to `{ repr: "<UNIMPLEMENTED>", kind: null }`.

PR #161, opened 2021-09-09
feat: add doc for infer, mapped and import types

kitsonk:
Closes: #134

In addition to mapped types in #134, I also added infer types and import types as well, which completes all the functionality of types that are supported by swc/TypeScript.

There are still a couple matches that were producing `<TODO>`.  Like the change of `<UNIMPLEMENTED>`, I have changed them to `[UNSUPPORTED]` as the `<` `>` can cause issues when doing SSR if you don't sanitise the values.


bartlomieju (approved):
LGTM, no comments

PR #165, opened 2021-09-09
chore: update wasm build to Rust 1.55

kitsonk:


bartlomieju (approved):
LGTM, should we look into automating this task?

kitsonk:
Yes, there is an issue for it in deno_graph, I will open one here. It is beyond my personal expertise though.

Issue #42, opened 2020-10-15
Doc should be able to extract "module level" JSDoc


bartlomieju:
```ts

/** This module is intended to be 
 * used with latest deno runtime.
 * Lorem ipsum.
 **/


/** hello */
function foo() {
}
```

Doc generator should be able to extract first JSDoc and treat it as a "module documentation" - showing it at the top of terminal output or pass it to JSON structure.

CC @ry

wperron:
Grabbing the first jsdoc comment arbitrarily might lead to some undesired behavior; for instance if a module is implemented in plain js and uses jsdoc to declare type definitions and that's what's at the top of the module, it probably shouldn't be interpreted as module-level doc. Maybe we should instead look at the [`@module`](https://jsdoc.app/tags-module.html) tag rather than simply the first jsdoc encountered.

ry:
> Maybe we should instead look at the @module tag rather than simply the first jsdoc encountered 

It's unfortunate that in JavaScript (and especially TypeScript) it often requires extra syntax, extra characters to do the right thing. It has the effect of making the language tedious. Smaller code is worse, you must add boilerplate to make it better. `@module` is boilerplate. (BTW I'm not questioning the necessity of it - I have no idea - just noting that correct is unfortunately often longer than incorrect.)

jeremyBanks:
I have seen `@fileoverview` most often, but apparently [`@file` is canonical JSoc](https://jsdoc.app/tags-file.html). It would be great if Deno Doc could support this in some form. I'd like to add a brief description at the top of some of my module docs. 🙂

vovacodes:
I'd like to revive this discussion, I'm happy to take on this issue but we as a community need to agree on how we define the module-level comment.

- There were some valid points mentioned against simply grabbing the first JSDoc in the file. 
- `@module` or `@file` seem quite "boilerplaty", I'd agree with @ry on that.
- JSDoc is generally not a very friendly format for long-form comments that may include markdown - you need to escape a bunch of characters.

There is a very good reason Rust adopted `///` and `//!` for doc comments, it's super easy to parse and markdown-friendly.
I know we have legacy (JSDoc) that has been working well enough for a lot of people for years but it doesn't look like there is a lot of existing code that uses module-level comments so we could simply embrace the most popular format.

Since we are effectively introducing a relatively new concept to the js doc comments, why wouldn't we just pick what works and battle-tested in another existing ecosystem namely `//!` from Rust. Among the benefits of such an approach:
[...]

PR #167, opened 2021-09-10
feat: support leading `@module` JSDoc blocks

kitsonk:
Closes: #42

This supports finding a leading JSDoc blocks that contain a `@module` tag and creating a `DocNode` of kind `ModuleDoc` that contains the value as the `js_doc`.

It will only ever parse out a maximum of one leading block that appears to be JSDoc and contains the `@module` tag.  Multiple leading comment blocks containing a `@module`, only the first one will be added and stand-alone JSDoc comment blocks are ignored if they don't contain `@module`.

The `DocPrinter` doesn't print out the `ModuleDoc` nodes, as the printer doesn't appear to print out JSDoc at all at the moment, but the calls are stubbed out to deal with this in the future if we wish.


src/parser.rs
```diff
@@ -661,6 +662,12 @@ impl<'a> DocParser<'a> {
 
     let mut is_ambient = true;
 
+    if let Some((js_doc, span)) = leading_js_doc(parsed_source) {
+      let doc_node =
+        DocNode::module_doc(get_location(parsed_source, span.lo), js_doc);
+      doc_entries.push(doc_node);
+    }
```
bartlomieju, line 669: I was confused at first when I saw it. Maybe `leading_js_doc` could use a better name suggesting that it extracts "module" tag?

bartlomieju (approved):
LGTM, I have only a small nitpick

src/parser.rs
kitsonk: I've renamed it and also provided a doc block to clarify.

PR #168, opened 2021-09-11
fix: handle class JSDoc when decorators are present

### LINKED ISSUE #159: JSDocs fail to be interpreted with decorators
```
@Decorator()
/**
 * JSDOC
 */
export class MyClass { 

}
```
Would succeed to add the JSDoc information to the output
but
```
/**
 * JSDOC
 */
@Decorator()
export class MyClass { 

}
```
will **not** succeed.

Both should be supported

kitsonk:
Fixes #159


bartlomieju (approved):
LGTM

PR #195, opened 2021-10-01
fix: allow multi-line tag docs

### LINKED ISSUE #194: All untagged paragraphs following a tag are ignored
Paragraphs following a tag will be ignored if they're untagged. This is common (leading `@deprecated`, etc)

```ts
/**
 * Paragraph 1
 *
 * @deprecated Paragraph 2
 *
 * Paragraph 3
 *
 * @param paramagraph 4
 */
export let a;
```
Paragraph 3 gets eaten:
```json
[
  {
    "kind": "variable",
    "name": "a",
    "location": {
      "filename": "file:///Users/robert/Workspaces/github.com/denoland/deno_doc/repro.ts",
      "line": 10,
      "col": 0
    },
    "jsDoc": {
      "doc": "Paragraph 1\n",
      "tags": [
        {
          "kind": "deprecated",
          "doc": "Paragraph 2"
        },
        {
          "kind": "param",
          "name": "paramagraph",
          "doc": "4"
        }
      ]
    },
    "variableDef": {
      "tsType": null,
      "kind": "let"
    }
  }
]
```
`jsDoc.doc` should be `"Paragraph 1\n\nParagraph 3\n"`

<details>
<summary>Example from builtins (Deno.copy)</summary>

```js
  /**
   * @deprecated Use `copy` from https://deno.land/std/io/util.ts instead.
   * `Deno.copy` will be removed in Deno 2.0.
   *
   * Copies from `src` to `dst` until either EOF (`null`) is read from `src` or
   * an error occurs. It resolves to the number of bytes copied or rejects with
   * the first error encountered while copying.
   *
   * ```ts
   * const source = await Deno.open("my_file.txt");
   * const bytesCopied1 = await Deno.copy(source, Deno.stdout);
   * const destination = await Deno.create("my_file_2.txt");
   * const bytesCopied2 = await Deno.copy(source, destination);
   * ```
   *
   * @param src The source to copy from
   * @param dst The destination to copy to
   * @param options Can be used to tune size of the buffer. Default size is 32kB
   */
```
Becomes this output. Notice the missing second paragraph and example - `jsDoc` should have a `doc` property.
```json
"kind": "function",
"name": "copy",
"location": {
  "filename": "deno://lib.deno.d.ts",
  "line": 441,
  "col": 2
},
"jsDoc": {
  "tags": [
    {
      "kind": "deprecated",
      "doc": "Use `copy` from https://deno.land/std/io/util.ts instead."
    },
    {
      "kind": "param",
      "name": "src",
      "doc": "The source to copy from"
    },
    {
      "kind": "param",
      "name": "dst",
      "doc": "The destination to copy to"
    },
    {
      "kind": "param",
      "name": "options",
      "doc": "Can be used to tune size of the buffer. Default size is 32kB"
    }
  ]
},
```

</details>

Found this while fixing denoland/doc_website#252 - this blocks upgrading the doc website's Deno version.

SeparateRecords:
This PR changes the tests to also check for multi-line support where docs are allowed.

Closes #194

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        os: [macOS-latest, ubuntu-16.04, windows-2019]
+        os: [macOS-latest, ubuntu-latest-xl, windows-2019]
 
     env:
       CARGO_INCREMENTAL: 0
--- a/lib/deno_doc.generated.js
+++ b/lib/deno_doc.generated.js
@@ -217,11 +217,11 @@ function isLikeNone(x) {
   return x === undefined || x === null;
 }
 /**
-* @param {string} root_specifier
-* @param {Function} load
-* @param {Function | undefined} maybe_resolve
-* @returns {Promise<any>}
-*/
+ * @param {string} root_specifier
+ * @param {Function} load
+ * @param {Function | undefined} maybe_resolve
+ * @returns {Promise<any>}
+ */
 export function doc(root_specifier, load, maybe_resolve) {
   var ptr0 = passStringToWasm0(
     root_specifier,
--- a/src/js_doc.rs
+++ b/src/js_doc.rs
@@ -5,17 +5,17 @@ use serde::Deserialize;
 use serde::Serialize;
 
 lazy_static! {
-  static ref JS_DOC_TAG_MAYBE_DOC_RE: Regex = Regex::new(r#"^\s*@(deprecated)(?:\s+(.+))?"#).unwrap();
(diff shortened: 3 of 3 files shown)


SeparateRecords:
Is CI broken or have I done something wrong?

bartlomieju:
> Is CI broken or have I done something wrong?

It seems it's broken, I changed the CI script to use `ubuntu-latest-xl` runner instead, let's see if this fixes the problem.

SeparateRecords:
I've formatted the generated JS file, but this PR hasn't touched it otherwise. Should I regenerate the JS+WASM or leave that for the next release to deal with?

bartlomieju (approved):
LGTM, thank you @SeparateRecords

PR #196, opened 2021-10-04
(feat) Support TS parameter properties

### LINKED ISSUE #192: Parameter properties are ignored
Typescript has [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties), but `deno_doc` ignores them. To reproduce, compare the output between running `deno doc` on this:

```typescript
class Example {
  name: string

  constructor(name: string) {
    this.name = name
  }
}
```

vs this

```typescript
class Example {
  constructor(
    public name: string
  ) {}
}
```

`deno doc` will output the `name` property for the first example but not for the second.

LionC:
Closes #192

### DIFF
--- a/src/class.rs
+++ b/src/class.rs
@@ -205,6 +205,35 @@ pub struct ClassDef {
   pub decorators: Vec<DecoratorDef>,
 }
 
+fn ts_param_prop_to_property_def(
+  parsed_source: &ParsedSource,
+  param_prop: &deno_ast::swc::ast::TsParamProp,
+) -> ClassPropertyDef {
+  use deno_ast::swc::ast::Pat;
+  use deno_ast::swc::ast::TsParamPropParam;
+
+  let ident = match &param_prop.param {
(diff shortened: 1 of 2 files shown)


kitsonk:
Thank you.  As I stated in https://github.com/denoland/deno_doc/issues/192#issuecomment-933300143 though, I don't think this should generate a class property node.  It should just be adding the accessibility value to the arguments on a constructor function.

LionC:
I though the purpose of `deno doc` is to document a module's API? Whether a property is a parameter property or not is an implementation detail - it is actually completely opaque during runtime, as param properties do not exist in JS, which would make the doc output weird to JS users as well.

The property being noted in a completely different place in the docs due to how the module author implemented it internally feels strange to me.

bartlomieju:
Please rebase against `main`, I've updated the CI script and it should make ubuntu step pass

kitsonk:
Thanks for this but closed in favour of #267

PR #212, opened 2021-11-23
feat: provide declaration kind and provide JS interface to private

kitsonk:
This PR does two things:

- Adds a `declaration_type` property to a `DocNode` which is set to the _reason_ why the doc node exists.
- Allowed the JavaScript/TypeScript interface to use the _private_ setting which includes all doc nodes/symbols in the output, not just those being exported.

This is designed to make it possible to generate documentation for everything in a module or ambient type declaration file and display that information usefully. This will allow us to build towards symbol references and scope analysis.


lib/types.d.ts
```diff
@@ -11,10 +11,13 @@ export type DocNode =
   | DocNodeInterface
   | DocNodeImport;
 
+export type DeclarationKind = "private" | "export" | "import" | "declaration";
```
lucacasonato, line 14: Having a comment here explaining what each of these mean would be very useful. Does `"declaration"` mean the `declare` keyword in TS?

lucacasonato (approved):

lib/types.d.ts
kitsonk: Hmmm... yeah, maybe `declare` is better, yes that is what it means.  I will change it to `declare` and add some comments.

lucacasonato:
Actually, one more comment:

Would it not make sense to remove the `import` kind and just mark imports as a `private` declaration? The distinction between `import` and `private` is not clear to me, because imports can already be found by looking for nodes with `kind: "import"`.

kitsonk:
That makes sense, will amend.

PR #213, opened 2021-11-28
fix: correct interface properties and methods

kitsonk:
A couple of tweaks of the interface properties and methods in the documents to properly represent them.


src/swc_util.rs
```diff
@@ -16,6 +16,10 @@ lazy_static! {
   static ref JS_DOC_RE: Regex = Regex::new(r#"\s*\* ?"#).unwrap();
 }
 
+pub(crate) fn is_false(b: &bool) -> bool {
+  !b
+}
```
lucacasonato, line 21: Oof.

lucacasonato (approved):
LGTM!

src/swc_util.rs
kitsonk: Is there a better way?

Just trying to strip out "useless" fields from the JSON serialization, as with large doc structures, it becomes a "problem".  `lib.dom.d.ts` for example is 9.9mb serialized. 😱

dsherret: I feel like we should use this and stuff like `skip_serializing_if = "Option::is_none"`  liberally one day.

src/interface.rs
```diff
@@ -9,6 +9,7 @@ use crate::js_doc::JsDoc;
 use crate::node::DeclarationKind;
 use crate::params::ts_fn_param_to_param_def;
 use crate::swc_util::get_location;
+use crate::swc_util::is_false;
```
dsherret, line 12: No big deal, but this isn't really related to swc.

dsherret (approved):
LGTM

src/swc_util.rs
lucacasonato: > Is there a better way?

No, I think this is right. Still pretty funny though 😅

src/interface.rs
kitsonk: There was no other good place to put it and I wasn't going to just create another module for it.

PR #217, opened 2021-12-14
chore: integrate API changes in deno_graph

kitsonk:
We should land and publish `deno_graph` before merging this.


src/parser.rs
```diff
@@ -136,11 +137,18 @@ impl<'a> DocParser<'a> {
...
+      self.parse_source(
+        &module.specifier,
+        module.media_type,
+        module.source.clone(),
+      )
+    } else {
+      Err(DocError::Resolve(format!(
+        "{} is not an JavaScript/TypeScript module",
+        specifier
+      )))
```
bartlomieju, line 150: Will this be a fatal error for https://doc.deno.land? If there's appropriate prompt then it's okay

bartlomieju (approved):

src/parser.rs
kitsonk: No, it is likely to continue to break at the build of the graph.  If you try to doc a JSON module today you get:

https___doc_deno_land_https___deno_land_std_0_117_0_deno_json

PR #222, opened 2021-12-20
fix: parse param decorators

### LINKED ISSUE #220: Support parameter decorators
A parameter decorator is a [thing](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators) but currently they aren't part of the doc structure, but should be.  I am not sure swc parses them out.

kitsonk:
Fixes #220


bartlomieju (approved):
LGTM

Issue #163, opened 2021-09-09
Infer simple properties types


kitsonk:
Currently if you have this:

```ts
export class A {
  a = "string";
  b = 100;
}
```

The property `a` and the property `b` will have no type associated with them.  We currently infer those types for `const` variables, but don't do it for properties.

PR #232, opened 2022-01-12
feat: infer simple property types, infer array types

kitsonk:
Closes: #163

Now the following:

```ts
export class A {
  a = "a";
  b = [1, "b"];
}
```

Will generate the equivalent doc that would be something like this:

```ts
export class A {
  a: string;
  b: (string | number)[];
}
```

In addition it fixes a few issues with inferring types I encountered and for arrays, it also supports const assertions:

```ts
export const a = [1, "a"];
export const b = [1, "b"] as const;
```

Would render as:

```ts
export const a: (number | string)[];
export const b: (1 | "b")[];
```


bartlomieju (approved):
LGTM, great improvement

PR #233, opened 2022-01-12
feat: infer function types for variables and properties

### LINKED ISSUE #162: Infer types variables and properties assigned functions
Currently, if you have:

```typescript
class A {
  a = (a: string): void => {}
}
/* or */
const a = (a: string): void => {}
```

There is no type information available in the output.  Inferring the type of the function should be fairly straight forward to infer.

kitsonk:
Closes #162

This supports inferring the function signature of properties and variables when a type isn't specified.  So for example the following:

```ts
export class A {
  a = (a:string): void => {};
}

export const af = (a: string): void => {};
export const fn = function (a: string): void {};
```

Would render something like:

```ts
export class A {
  a: (a:string) => void;
}

export const af: (a: string) => void;
export const fn: (a: string) => void;
```


bartlomieju (approved):
LGTM

PR #235, opened 2022-01-15
reafactor: build type only graphs for docs

kitsonk:
This will improve the "performance" of large graphs that contain a lot of JavaScript code, like `dnt`, specifically to allow docland to render these large graphs.


bartlomieju (approved):
LGTM

PR #243, opened 2022-02-03
chore: update WASM build

kitsonk:
PR #242 was merged without doing a build (which also means the published version on `deno.land/x` doesn't have the right dependencies in the build)


bartlomieju (approved):
LGTM, my bad

PR #255, opened 2022-07-05
feat: JavaScript interface supports import maps

kitsonk:


bartlomieju (approved):
LGTM!

src/js.rs
```diff
@@ -93,18 +127,53 @@ pub async fn doc(
...
   console_error_panic_hook::set_once();
   let root_specifier = ModuleSpecifier::parse(&root_specifier)
     .map_err(|err| JsValue::from(js_sys::Error::new(&err.to_string())))?;
   let mut loader = JsLoader::new(load);
-  let maybe_resolver = maybe_resolve.map(JsResolver::new);
+  let maybe_resolver: Option<Box<dyn Resolver>> = if let Some(import_map) =
+    maybe_import_map
+  {
+    if maybe_resolve.is_some() {
+      console_warn!("An import map is specified as well as a resolve function, ignoring resolve function.");
```
dsherret, line 140: Maybe something for future improvement: maybe we should return all these diagnostics instead of logging them to the console? Overall, probably doesn't matter for the use cases of this repo.

dsherret (approved):
Welcome back! LGTM

src/js.rs
kitsonk: Return them to where?  I think we should just keep the return of `doc()` clean.  If it is fatal it throws.

PR #262, opened 2022-07-20
fix: add has_body to function definitions and constructors

### LINKED ISSUE #259: Deno doc shouldn't include the implementation signature of functions and methods
Given a method or function that looks like the following:

```ts
// example.ts
export function env(items: Record<string, string | undefined>): void;
export function env(name: string, value: string | undefined): void;
export function env(nameOrItems: string | Record<string, string | undefined>, value?: string) {
}
```

Running:

```
deno doc example.ts
```

Outputs:

```
> deno doc example.ts       
Defined in file:///V:/dax/example.ts:1:0

function env(items: Record<string, string | undefined>): void

Defined in file:///V:/dax/example.ts:2:0

function env(name: string, value: string | undefined): void

Defined in file:///V:/dax/example.ts:3:0

function env(nameOrItems: string | Record<string, string | undefined>, value?: string)
```

This shouldn't happen because the implementation signature is an implementation detail that the consumer should be concerned about. For example, TS's language server does not display the implementation signature:

image

kitsonk:
This adds `has_body` to constructors and function definitions and only print out overloads, skipping implementations for class methods, functions and constructors.  In additional, constructors can be, surprisingly optional, but was not part of the doc node structure.

Fixes #259


dsherret (approved):
LGTM other than the failing test (which on very brief glance seems to be some unstable ordering).

kitsonk:
It isn't unstable ordering (the diff library just highlights unstable ordering) it is that one of the items used in the new tests had other changes in other PR and when I rebased, I didn't update the test.  Will do so now.

PR #267, opened 2022-07-26
feat: support TypeScript constructor params

### LINKED ISSUE #192: Parameter properties are ignored

kitsonk:
Closes #192

This is an alternative to #196. Like the recent has_body change, instead of inferring, the doc nodes returned more reflect the code as written, and it is up to the printers to determine how to display that to the user.

In this situation, printers need to decide how to display a `public` constructor param, along with the other accessibility keywords, override and readonly, which is to likely reflect them in the constructor signature as well as the properties documented on the class.


bartlomieju (approved):
LGTM

lib/types.d.ts
```diff
@@ -95,13 +95,19 @@ export interface ClassDef {
   decorators?: DecoratorDef[];
 }
 
+export type ClassConstructorParamDef = ParamDef & {
```
dsherret, line 98: Nitpick: seems more consistent to use an interface?

dsherret (approved):
LGTM too

lib/types.d.ts
kitsonk: You can't here, because `ParamDef` is a tagged union type alias and interfaces can't extend type aliases.

PR #283, opened 2022-08-18
refactor: upgrade for deno_graph 0.31

dsherret:
Waiting on https://github.com/denoland/deno_graph/pull/173


benches/doc_parser.rs
```diff
@@ -24,19 +25,19 @@ async fn parse_with_reexports() -> Vec<DocNode> {
...
+  let analyzer = RefCellCapturingParsedSourceAnalyzer::default();
   let graph = create_type_graph(
     vec![root.clone()],
     false,
     None,
     &mut memory_loader,
     None,
     None,
-    None,
+    Some(&analyzer),
```
kitsonk, line 36: Seeing this in action raises a few questions in my mind.

Why do we have to pass an analyzer as a consumer? Should this just be handled for me. If I want to change default behaviour sure, but the default behaviour should just work without passing in stuff.

Second, `RefCellCapturingParsedSourceAnalyzer` is like `getElementsById`, accurately named but unnecessarily long. It assumes there are other varieties of analyzers that aren't refcell or aren't parsed source. Assume this is the most common type of analyzer, maybe just analyzer and other variants can denote their distinction, if necessary. The name doesn't impart any real meaning about it, especially because it is just default configured.

dsherret: Generally someone doesn't need to bother with this and they can just pass `None`. It's only if they want to re-use the source files or provide a different kind of analyzer.

Yeah, `RefCellCapturingParsedSourceAnalyzer` was a bad name you saw I asked about and thanks for the new suggestion. I've broken this all up now.

bartlomieju (approved):
LGTM

PR #292, opened 2022-09-18
fix: use capturing parser instead of store in order to parse when ParsedSource is not in store

dsherret:
Same fix as https://github.com/denoland/eszip/pull/88/, but for deno_doc.


kitsonk (approved):
LGTM

PR #293, opened 2022-09-20
feat: support destructured var exports

kitsonk:


src/tests.rs
```diff
@@ -2236,6 +2236,95 @@ export const tpl2 = `Value: ${num}`;
...
+            "repr": "string",
+            "kind": "keyword",
+            "keyword": "string"
+          },
+          "kind": "const"
+        }
+      },
+      {
+        "kind": "variable",
+        "name": "h",
```
kitsonk, line 2309: @crowlKats this is failing, as the exported symbol name should be `h` but it is currently being set to `g`.

src/tests.rs
```diff
@@ -2236,6 +2236,95 @@ export const tpl2 = `Value: ${num}`;
     ]
   );
 
+  json_test!(export_const_destructured,
+    r#"const c = { a: "a", b: 2 };
+    const d: { f: string; g: number; } = { f: "f", g: 2 };
+
+    export const { a, b }: {
+      /** type alias doc */
```
kitsonk, line 2244: The tsc language server associates JSDoc in the type with the variable, so I would expect `a` and `b` to have JSDoc associated with them.  I haven't added it to the fixture yet.

kitsonk (commented):
@crowlKats I have added a test case and could only find two things "wrong" with it:

- The binding name on the export should be correct.  In the test case, `g` is exported as `h`.
- We should associate JSDoc from the types with the exported symbols where possible, as the TypeScript language server does.

crowlKats (approved):

PR #295, opened 2022-09-23
feat: infer object literals

crowlKats:


src/tests.rs
```diff
@@ -4334,6 +4334,164 @@ export function foo(bar: any): asserts bar {}
...
+          "col": 0
+        },
+        "declarationKind": "export",
+        "variableDef": {
+          "tsType": {
+            "repr": "",
+            "kind": "typeLiteral",
+            "typeLiteral": {
+              "methods": [],
+              "properties": [
```
kitsonk, line 4367: The first challenge is that this appears to not be deterministic, but other situations are.  Having it not be deterministic I think is going to cause problems.  Like when we generate anchors, we would expect it to be consistent ordering, which when declared it obviously has consistent ordering.  I am not sure offhand why it is not deterministic here.

src/tests.rs
```diff
@@ -4334,6 +4334,164 @@ export function foo(bar: any): asserts bar {}
...
+                  "optional": false,
+                  "tsType": {
+                    "repr": "string",
+                    "kind": "keyword",
+                    "keyword": "string"
+                  },
+                  "typeParams": []
+                },
+                {
+                  "name": "b",
```
kitsonk, line 4381: This inferred better than I expected.

kitsonk (commented):
Some thoughts...

src/tests.rs
kitsonk: Used IndexMap instead of HashMap.

kitsonk (approved):
LGTM

PR #300, opened 2022-10-16
feat(jsdoc): default & optional parameter value

crowlKats:
as per last case of https://jsdoc.app/tags-param.html#optional-parameters-and-default-values.
This came up in https://github.com/denoland/deno/pull/16299/files#r996347998

One thing thats missing is the ability to mark something as optional; unsure how that could be structured, as a double option for `default` would be great though serde doesnt handle that (would need to use an additional crate)


kitsonk (approved):
LGTM!

kitsonk (approved):
LGTM

PR #309, opened 2022-11-23
feat: support re-exported json modules

dsherret:


src/parser.rs
```diff
@@ -271,6 +272,11 @@ impl<'a> DocParser<'a> {
       module
     };
 
+    if module.kind == ModuleKind::Asserted {
+      // for now, don't display anything for json modules
+      return Ok(Vec::new());
```
dsherret, line 277: Made this PR a draft again. Going to make this return a doc node while I'm here.

crowlKats (approved):

PR #323, opened 2023-02-27
feat: handle @ignore tags internally

crowlKats:
unsure if we should remove the typings: the api will never give ignore tags, but the rust API of deno_doc will still have it part of the JsDocTag enum, so it would mismatch. Maybe no-doc the Ignore tag?

### DIFF
--- a/src/class.rs
+++ b/src/class.rs
@@ -10,6 +10,7 @@ use crate::decorators::DecoratorDef;
 use crate::function::function_to_function_def;
 use crate::function::FunctionDef;
 use crate::js_doc::JsDoc;
+use crate::js_doc::JsDocTag;
 use crate::node::DeclarationKind;
 use crate::params::assign_pat_to_param_def;
 use crate::params::ident_to_param_def;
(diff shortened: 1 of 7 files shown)


src/class.rs
```diff
@@ -432,9 +439,12 @@ pub fn class_to_class_def(
   let js_doc = if !class.decorators.is_empty() {
     js_doc_for_range(parsed_source, &class.decorators[0].range())
   } else {
-    JsDoc::default()
+    Some(JsDoc::default())
   };
 
+  // TODO
+  let js_doc = js_doc.unwrap_or_default();
+
```
crowlKats, line 447: todo

src/parser.rs
```diff
@@ -788,6 +830,8 @@ impl<'a> DocParser<'a> {
       doc_entries.push(doc_node);
     }
 
+    // TODO: if module doc has an Ignore tag, we need to ignore the entire module's contentx
```
crowlKats, line 833: todo

dsherret (approved):
LGTM!

PR #325, opened 2023-03-14
fix: walk down member expressions when resolving class extensions

### LINKED ISSUE #324: Class extends does not work with `obj.Class`
**Behavior**

Given the following typescript file...

```ts
class Foo {}
const obj = { Foo }

export class Bar extends obj.Foo {}
```

... `deno_doc` will output the following tree.

```json
[
  {
    "kind": "class",
    "name": "Bar",
    "location": {
      "filename": "file:///some_file.ts",
      "line": 4,
      "col": 0
    },
    "declarationKind": "export",
    "classDef": {
      "isAbstract": false,
      "constructors": [],
      "properties": [],
      "indexSignatures": [],
      "methods": [],
      "extends": null,
      "implements": [],
      "typeParams": [],
      "superTypeParams": []
    }
  }
]
```

Note that `classDef.extends` is `null`.

**Expected Behavior**

I would expect `classDef.extends` to be populated with the string `"obj.Foo"`, or, if not that, at least `"Foo"`.

Related to #136

Vehmloewff:
Fixes #324


crowlKats:
Could you please rebase?

Vehmloewff:
Sure thing

Vehmloewff:
_ping :)_

crowlKats:
there seem to be a lot of unrelated changes

Vehmloewff:
It is because I rebased instead of merging, which is, I believe, what you wanted me to do.

dsherret (approved):
LGTM. Thanks @Vehmloewff!

In the future, if you see a bunch of unrelated changes in the PR after rebasing or merging, then something probably went wrong. Also, I'd recommend just merging as it doesn't require force pushing and we squash the commits in the PR anyway.

PR #329, opened 2023-03-25
feat: add support of @example name

JOTSR:
Support name of example if specified.
Parse <caption>$name</caption> for @example tag as specified in jsdoc.

example:
```ts
/**
 * Asynchronous sleep during the specified delay.
 * @example
 * ```ts
 * await sleep(500)
 * //wait 500ms
 * ```
 * @example <caption>Use abort controller</caption>
 * ```ts
 * const ac = new AbortController()
 * await sleep(1_000_000, ac)
 * //ac called elsewhere, eg: callback
 * ac.abort()
 * //abort sleep before long sleep
 * ```
 * @param {number} delay - The amount of time to wait in milliseconds.
 * @param {{signal: AbortSignal}=} { signal } - Optional abort controller signal to stop sleeping.
 */
export function sleep(
	delay: ms,
	{ signal }: { signal?: AbortSignal } = {},
): Promise<void> {
	return new Promise((resolve, reject) => {
		const timer = setTimeout(resolve, delay)

		signal?.addEventListener('abort', () => {
			clearTimeout(timer)
			reject(signal.reason)
		}, {
			once: true,
		})
	})
}
```
 will now output doc
```sh
function sleep(delay: ms, {signal}: { signal: AbortSignal; }): Promise<void>
  Asynchronous sleep during the specified delay.

  @example Async sleep
      ```ts
      await sleep(500)
      //wait 500ms
      ```

  @example
      ```ts
      const ac = new AbortController()
      await sleep(1_000_000, ac)
      //ac called elsewhere, eg: callback
      ac.abort()
      //abort sleep before long sleep
      ```

  @param {number} delay
      - The amount of time to wait in milliseconds.

  @@param {{signal: AbortSignal}=} { signal } - Optional abort controller signal to stop sleeping.
```


crowlKats:
doc_components currently gets the name from the first part (everything before the first empty line or before a code block) of the doc content of the example tag. I don't understand why this is necessary

JOTSR:
Which doc_component do you refers. The first goal is to patch the name in example as required by jsdoc spec and then align to other tools (like jedoc code hint in vscode). As describe in PR (and spec) it gives context to example before trying to read it in detail. The second part is to fix the issues in deno doc api consumers, for instance module doc on deno.land get blank example name when using jsdoc synthax for example naming. To align with others tools and don't throw a spec option I fixed this lack of implementation. Then I will, if PR is accepted, fix doc render on deno.land or if not, add an issue to compensate the api output.

PR #350, opened 2023-09-27
feat(js): add printImportMapDiagnostics option

crowlKats:


src/js.rs
```diff
@@ -130,7 +131,7 @@ pub async fn doc(
   let maybe_resolver: Option<Box<dyn Resolver>> = if let Some(import_map) =
     maybe_import_map
   {
-    if maybe_resolve.is_some() {
+    if print_import_map_diagnostics && maybe_resolve.is_some() {
```
dsherret, line 134: What might be better is to instead return the import map diagnostics. Then someone can choose to print them or do whatever. This seems good enough though.

dsherret (approved):
LGTM

src/js.rs
crowlKats: I like that, but dont really want to change the return value of the doc function. Let's do this for now and if necessary revisit in the future

PR #357, opened 2023-10-25
fix: properly display unsupported tags

roj1512:
**Before**

image_2023-10-25_21-04-01

**Now**

image


src/printer.rs
```diff
@@ -373,7 +373,11 @@ impl<'a> DocPrinter<'a> {
         self.format_jsdoc_tag_maybe_doc(w, doc, indent)
       }
       JsDocTag::Unsupported { value } => {
-        writeln!(w, "{}@{}", Indent(indent), colors::magenta(value))
+        let mut name =
+          String::from(value.split_whitespace().next().unwrap_or("@"));
+        name.remove(0);
+        let value = &value[name.len() + 1..];
```
dsherret, line 379: Could you write a test to show this won't panic when the value only has a name?

dsherret (changes requested):

src/printer.rs
dsherret: Thanks! You might have to merge main in to get the CI to pass.

roj1512: OK.

dsherret: In the future I'd recommend not opening PRs from a main branch because then I can't push to fix the compile error and get this quickly merged. Can you cherry pick this commit to your branch? https://github.com/dsherret/deno_doc/commit/ca7ddcd2fffa9e81a391a4ce1ad69cee9647dd90

roj1512: I didn’t know about that. Done.

dsherret (approved):
LGTM. Thanks a lot @roj1512!

PR #358, opened 2023-10-25
feat: rewrite parser to use deno_graph's type tracing

dsherret:
This is a rewrite of deno_doc's parser to use deno_graph's type tracing. This makes deno doc way smarter at analyzing types because it uses a symbol graph.

Closes https://github.com/denoland/deno_doc/issues/340


examples/ddoc/main.rs
dsherret, line 1: I forgot I removed this. I will see about adding it back.

src/node.rs
```diff
@@ -5,6 +5,11 @@ use serde::Serialize;
 
 use crate::js_doc::JsDoc;
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct NamespaceDef {
+  pub elements: Vec<DocNode>,
+}
```
dsherret, line 11: Didn't seem worth it to have a separate file for just this... actually we may want to consider collapsing many of these modules into here and parser.rs.

src/tests.rs
```diff
@@ -591,15 +637,15 @@ export namespace Deno {
...
   let entries = DocParser::new(graph, false, analyzer.as_capturing_parser())
+    .unwrap()
     .parse(&specifier)
     .unwrap();
 
   // Namespace
   let found =
     find_nodes_by_name_recursively(entries.clone(), "Deno".to_string());
-  assert_eq!(found.len(), 2);
+  assert_eq!(found.len(), 1);
```
dsherret, line 647: It's declaration merged now.

src/tests.rs
```diff
@@ -1151,7 +1200,7 @@ declare namespace RootNs {
           "location": {
             "filename": "file:///test.ts",
             "line": 4,
-            "col": 4
+            "col": 18
```
dsherret, line 1203: The variable declaration starts are now actually at the variable declaration starts instead of the variable statement.

src/tests.rs
```diff
@@ -810,7 +856,7 @@ async fn exports_imported_earlier_default() {
       "location": {
         "filename": "file:///foo.ts",
         "line": 1,
-        "col": 0
+        "col": 6
```
dsherret, line 859: The variable declaration starts are now actually at the variable declaration starts instead of the variable statement.

src/variable.rs
```diff
@@ -17,127 +19,116 @@ pub struct VariableDef {
...
+    deno_ast::swc::ast::Pat::Object(pat) => {
+      for prop in &pat.props {
+        let (name, reassign_name, maybe_range) = match prop {
+          deno_ast::swc::ast::ObjectPatProp::KeyValue(kv) => (
+            crate::params::prop_name_to_string(None, &kv.key),
+            match &*kv.value {
+              deno_ast::swc::ast::Pat::Ident(ident) => {
+                Some(ident.sym.to_string())
+              }
+              _ => None, // TODO: properly implement
```
dsherret, line 100: These are todos from before.

src/js.rs
```diff
@@ -133,26 +133,42 @@ pub async fn doc(
...
-  let root_specifier = ModuleSpecifier::parse(&root_specifier)
-    .map_err(|err| JsValue::from(js_sys::Error::new(&err.to_string())))?;
+  inner_doc(
+    root_specifier,
+    include_all,
+    load,
+    maybe_resolve,
+    maybe_import_map,
+    print_import_map_diagnostics,
+  )
```
bartlomieju, line 143: Maybe we should box this future to avoid stack overflows on Windows? Though I'm not sure we have a codepath in CLI that would hit it

src/parser.rs
```diff
@@ -20,25 +21,42 @@ use crate::ImportDef;
...
+use deno_ast::swc::ast::TsEnumDecl;
+use deno_ast::swc::ast::TsInterfaceDecl;
+use deno_ast::swc::ast::TsModuleDecl;
+use deno_ast::swc::ast::TsModuleName;
+use deno_ast::swc::ast::TsTypeAliasDecl;
+use deno_ast::swc::ast::VarDecl;
 use deno_ast::swc::ast::VarDeclKind;
+use deno_ast::swc::ast::VarDeclarator;
 use deno_ast::ParsedSource;
+use deno_ast::SourceRange;
```
bartlomieju, line 45: So many of these... Maybe we should qualify them as `ast::FnDecl`, etc? Can be done in a follow up as it's a mechanical refactor

src/parser.rs
```diff
@@ -85,19 +103,39 @@ pub struct DocParser<'a> {
...
   pub fn new(
     graph: ModuleGraph,
     private: bool,
     parser: CapturingModuleParser<'a>,
-  ) -> Self {
-    DocParser {
+  ) -> Result<Self, anyhow::Error> {
+    struct NullTypeTraceHandler;
+
+    // todo(dsherret): surface these diagnostics
```
bartlomieju, line 117: What does this error mean?

src/parser.rs
```diff
@@ -106,49 +144,27 @@ impl<'a> DocParser<'a> {
...
-    let import_doc_entries = self.get_doc_nodes_for_module_imports(
-      &parsed_source,
-      module.body.clone(),
-      specifier,
-    )?;
-    definitions.extend(import_doc_entries);
-    let reexports = self.get_reexports_for_module_body(&module.body);
+    let module_symbol = self.get_module_symbol(specifier)?;
+    let definitions = self.get_doc_nodes_for_module_symbol(module_symbol)?;
+    let reexports = self.get_reexports_for_module(module_symbol);
```
bartlomieju, line 149: This is lovely

src/parser.rs
```diff
@@ -777,163 +872,208 @@ impl<'a> DocParser<'a> {
...
+    &self,
+    module_symbol: ModuleSymbolRef,
+    symbol: &Symbol,
+    maybe_node: Option<SymbolNodeRef<'_>>,
+  ) -> Option<DocNode> {
+    match module_symbol {
+      ModuleSymbolRef::Json(module_symbol) => parse_json_module_doc_node(
+        module_symbol.specifier(),
+        module_symbol.text_info().text_str(),
+      ),
```
bartlomieju, line 977: I don't understand this bit - how can JSON have documentation?

bartlomieju (approved):

src/parser.rs
dsherret: I'm going to remove this todo. Right now it provides diagnostics for when certain default expressions can't be analyzed.

src/parser.rs
dsherret: JSON modules can be imported then exported as an object. I see there's no test for this so I'll add one.

src/js.rs
dsherret: This is for the wasm code, so won't affect windows.

src/parser.rs
dsherret: Oh, nevermind. There is a test.

PR #359, opened 2023-10-26
feat: print visibility specifier before a symbol

bartlomieju:


src/node.rs
```diff
@@ -47,6 +47,18 @@ pub enum ReexportKind {
...
+  /// ```ignore
+  /// class Other {}
+  ///
+  /// export class Test {
+  ///     foo: string;
+  /// }
+  /// ```
+  ///
+  /// In the above example `Other` would have `implicit` equal to true.
+  pub implicit: bool,
```
dsherret, line 61: Do we need this? Where is it used?

dsherret (approved):
LGTM!

PR #361, opened 2023-10-26
feat: display non-exported types referenced in exported types

dsherret:
Closes https://github.com/denoland/deno_doc/issues/277

### DIFF
--- a/js/test.ts
+++ b/js/test.ts
@@ -13,7 +13,7 @@ Deno.test({
     const entries = await doc(
       "https://deno.land/std@0.104.0/fmt/colors.ts",
     );
-    assertEquals(entries.length, 48);
+    assertEquals(entries.length, 49);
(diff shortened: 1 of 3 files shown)


js/test.ts
```diff
@@ -13,7 +13,7 @@ Deno.test({
     const entries = await doc(
       "https://deno.land/std@0.104.0/fmt/colors.ts",
     );
-    assertEquals(entries.length, 48);
+    assertEquals(entries.length, 49);
```
dsherret, line 16: `Rgb` types is used in a parameter, but not exported: https://deno.land/std@0.104.0/fmt/colors.ts?source=

bartlomieju (approved):
LGTM, cool, this is useful. I wonder if we have any problems like that in `lib.deno.ns.d.ts`

PR #363, opened 2023-10-30
feat: HTML output, cont.

bartlomieju:
- various CSS fixes
- fuzzy search using fuse.js


bartlomieju:
We should improve the `<head>` section to contain more relevant tags.

crowlKats (approved):

PR #365, opened 2023-10-30
feat: add missing jsdoc and location to class' TsIndexSignature

dsherret:


src/tests.rs
```diff
@@ -1942,7 +1972,7 @@ export class Foobar extends Fizz implements Buzz, Aldrin {
           },
           "location": {
             "filename": "file:///test.ts",
-            "line": 27,
+            "line": 32,
```
dsherret, line 1975: It's extremely painful to update these expectations.

bartlomieju (approved):
LGTM 👍

PR #367, opened 2023-10-30
chore: rewrite json and contains tests to use spec files

dsherret:
Now we can easily update these tests by running with `UPDATE=1`.


tests/helpers/mod.rs
```diff
@@ -0,0 +1,187 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
```
dsherret, line 1: The code in here is mostly for deno_graph.

tests/specs/AbstractClass.txt
dsherret, line 1: I programmatically created these files from the past tests.

bartlomieju (approved):

PR #368, opened 2023-10-31
feat: add MissingJsDoc and MissingTypeRef diagnostics

dsherret:
It's not perfect, but probably good enough for the first pass. We should infer return types like `void` when there's no return statement.


src/diagnostics.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+  }
+}
+
+struct DiagnosticDocNodeVisitor<'a> {
+  diagnostics: &'a mut DiagnosticsCollector,
+}
+
+impl<'a> DiagnosticDocNodeVisitor<'a> {
+  pub fn visit_doc_nodes(&mut self, doc_nodes: &[DocNode]) {
```
bartlomieju, line 115: So it's pretty much a lint rule? 😅

src/tests.rs
```diff
@@ -749,10 +786,14 @@ async fn exports_imported_earlier_default() {
...
-    .parse_with_reexports(&specifier)
-    .unwrap();
+  let entries = DocParser::new(
+    &graph,
+    analyzer.as_capturing_parser(),
+    DocParserOptions::default(),
+  )
+  .unwrap()
+  .parse_with_reexports(&specifier)
+  .unwrap();
```
bartlomieju, line 796: Something for a follow up - add a helper that will set it all up and hide `unwrap`s

bartlomieju (approved):
Fantastic!

src/tests.rs
dsherret: Yeah, I think a lot of these tests could be moved to the specs folder

src/diagnostics.rs
dsherret: Yes, but much smarter than something like https://github.com/denoland/deno_lint/pull/972

bartlomieju: Should we close that issue because of this feature?

dsherret: Yeah, I'll have the CLI feature PR close it.

PR #369, opened 2023-10-31
feat: ignore diagnostics on nodes with `@internal` tag

dsherret:
This supresses diagnostics on nodes marked with `@internal`. This is something we can change in the future, but I think it makes sense.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -1,9 +1,12 @@
 // Copyright 2020-2023 the Deno authors. All rights reserved. MIT license.
 
+use crate::js_doc::JsDoc;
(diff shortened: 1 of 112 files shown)


src/diagnostics.rs
```diff
@@ -20,7 +23,8 @@ use std::fmt;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum DocDiagnosticKind {
   MissingJsDoc,
-  MissingTypeRef,
+  MissingExplicitType,
+  MissingReturnType,
```
dsherret, line 27: Also makes this diagnostic a bit more specific.

bartlomieju (approved):

PR #382, opened 2023-10-31
fix: sometimes missing some private nodes when has reexports

dsherret:


src/parser.rs
```diff
@@ -996,7 +999,7 @@ impl<'a> DocParser<'a> {
 
     let is_ambient = exports.is_empty() && !module_has_import(module_symbol);
     for child_id in module_symbol.child_decls() {
-      if !handled_symbols.insert(child_id) {
+      if !handled_symbols.insert((module_symbol.module_id(), child_id)) {
```
dsherret, line 1002: This would obviously ignore nodes with the same child id, but in different modules.

bartlomieju (approved):

PR #383, opened 2023-10-31
feat: infer void return types more

dsherret:


src/function.rs
```diff
@@ -65,3 +89,57 @@ pub fn get_doc_for_fn_decl(
   let fn_def = function_to_function_def(parsed_source, &fn_decl.function);
   (name, fn_def)
 }
+
+fn get_return_stmt_from_function(
```
dsherret, line 93: Lifted from my code here: https://github.com/dsherret/libpack/blob/cd66da8fda60e55a833ce843ece439603fed1ed5/rs_lib/src/dts.rs#L1407

bartlomieju (approved):
LGTM, would it possible to do the same for other "simple types"? Eg. you demoed this feature today and had two constants, something like:
```
export const SEP = isWindows ? "\\" : "/";
export const FOO = isWindows ? /adsf/ : /qwer/
```
Probably doable but requires quite a bit of work?

dsherret:
Yeah, doable. Let's look into it for the future.

PR #385, opened 2023-11-01
feat: better non-exported type diagnostic

dsherret:


src/parser.rs
```diff
@@ -1271,3 +1289,114 @@ fn definition_location(
...
+  /// Symbol identifiers that are not exported, but are referenced
+  /// by exported symbols.
+  non_exported_public_ids: HashSet<UniqueSymbolId>,
+}
+
+impl SymbolVisibility {
+  pub fn build(
+    graph: &ModuleGraph,
+    root_symbol: &deno_graph::type_tracer::RootSymbol,
+  ) -> Result<Self, DocError> {
```
dsherret, line 1304: I realized while working on this that the "type tracing" feature in deno_graph is somewhat useless and no longer used. We should refactor it to a general purpose symbol graph: https://github.com/denoland/deno_graph/issues/316

src/parser.rs
```diff
@@ -1061,7 +1015,26 @@ impl<'a> DocParser<'a> {
...
+            let mut diagnostics = diagnostics.borrow_mut();
+            for dep in deps {
+              let module =
+                self.root_symbol.get_module_from_id(dep.module_id).unwrap();
+              let symbol = module.symbol(dep.symbol_id).unwrap();
+              diagnostics
+                .add_private_type_in_public(doc, doc_id, module, symbol);
+            }
+          }
+        }
```
bartlomieju, line 1034: Considering use `let/else` here?

src/parser.rs
bartlomieju: Agreed

bartlomieju (approved):

src/parser.rs
dsherret: I'll extract it to a method and do so.

PR #387, opened 2023-11-01
feat: raise diagnostics on member instead of parent declaration

dsherret:


tests/specs/InternalTag.txt
```diff
@@ -62,12 +62,6 @@ class Other
 
     @internal
 
-Defined in file:///mod.ts:4:1
-
-private type PrivateType = string
-
-  @internal
```
dsherret, line 69: Bug.

src/util/swc.rs
```diff
@@ -115,3 +115,8 @@ pub fn module_export_name_value(
     ModuleExportName::Str(str) => str.value.to_string(),
   }
 }
+
+/// If the jsdoc has an `@internal` or `@ignore` tag.
+pub fn has_ignorable_js_doc_tag(js_doc: &JsDoc) -> bool {
+  js_doc.tags.iter().any(|t| matches!(t, JsDocTag::Ignore) || matches!(t, JsDocTag::Unsupported { value } if value == "@internal" || value.starts_with("@internal ")))
+}
```
bartlomieju, line 122: Why isn't it line-wrapping here?

src/visibility.rs
```diff
@@ -0,0 +1,226 @@
...
+      let decl_deps = symbol
+        .decls()
+        .iter()
+        .filter(|d| !d.has_overloads())
+        .flat_map(|decl| decl.deps().into_iter().map(move |dep| (decl, dep)))
+        .map(|(decl, dep)| (symbol, decl, dep))
+        .chain(
+          symbol
+            .members()
+            .iter()
```
bartlomieju, line 94: This function is really complex, would it make sense to unit test it?

bartlomieju (approved):
LGTM, just a couple comments, but this is a great feature!

src/visibility.rs
dsherret: It's captured by the integration tests.

src/util/swc.rs
dsherret: No idea. Rustfmt sometimes works in mysterious ways 🤷‍♂️

PR #396, opened 2023-11-10
feat: PrivateTypeRef - add reference type location when in other module

dsherret:
This is two commits. One for the feature messaged in the title, and the other for shortening the jsdoc comment message.


src/diagnostics.rs
```diff
@@ -19,46 +19,62 @@ use deno_graph::symbols::ModuleInfoRef;
...
+  pub location: Location,
+  pub kind: DocDiagnosticKind,
+}
+
+impl DocDiagnostic {
+  pub fn message(&self) -> Cow<str> {
+    match &self.kind {
       DocDiagnosticKind::MissingJsDoc => {
-        f.write_str("Missing JS documentation comment.")
+        Cow::Borrowed("Missing JSDoc comment.")
```
bartlomieju, line 48: This is better 👍

bartlomieju (approved):
LGTM 🎉

PR #398, opened 2023-11-16
fix: issue discovering re-exports of re-exports

dsherret:
I recently introduced a bug to deno_doc related to discovering re-exports. We didn't have a test so it wasn't caught.

### DIFF
--- a/tests/specs/ReExportAll.txt
+++ b/tests/specs/ReExportAll.txt
@@ -1,0 +1,78 @@
+# a.ts
+const test1: string = "";
+const test2: number = 3, test3: number = 3;
+const test4: number;
+const test5: number;
+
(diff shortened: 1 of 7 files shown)


tests/specs/ReExportAll.txt
```diff
@@ -0,0 +1,78 @@
...
+# diagnostics
+[
+  "file:///b.ts:1:1 Missing JSDoc comment.",
+  "file:///a.ts:1:7 Missing JSDoc comment."
+]
+
+# output.txt
+Defined in file:///a.ts:1:7
+
+const test1: string
```
dsherret, line 29: This previously wasn't showing up.

bartlomieju (approved):
LGTM

PR #402, opened 2023-11-29
fix: proper single file mode handling

crowlKats:


dsherret (changes requested):
Considering this was a regression, we should spend the time adding some tests for this.

PR #403, opened 2023-11-30
fix: don't report diagnostics on remote modules

dsherret:
For https://github.com/denoland/deno/issues/21390


bartlomieju (approved):
LGTM

PR #404, opened 2023-11-30
feat(lint): don't require jsdoc for ctors with no parameters

dsherret:
It's not worth documenting a ctor that has no parameters; however, once it has parameters it's still good to document in order to describe the parameters. In case they don't want to bother describing the parameters, people can write a generic `/** Constructs a new instance. */` jsdoc to supress this lint.

https://github.com/denoland/deno/issues/21405


bartlomieju (approved):
LGTM, very nice 👍

PR #410, opened 2023-12-05
chore: fix deno_doc html tests on windows

dsherret:
The tests weren't passing on Windows:

https://github.com/denoland/deno_doc/actions/runs/7100208969/job/19325804254

It should now properly fail due to using bash for the shell (terrible GitHub actions decision to not do that by default).


Cargo.toml
```diff
@@ -47,7 +47,7 @@ tokio = { version = "1.25.0", features = ["full"] }
 pretty_assertions = "1.0.0"
 
 [features]
-default = ["rust"]
+default = ["html", "rust"]
```
dsherret, line 50: Otherwise you have to remember to include it during development. We might as well just include it all the time.

dsherret:
Waiting on https://github.com/denoland/wasmbuild/pull/111 -- Seems like wasmbuild was broken for `--no-default-features`

lib/Cargo.toml
dsherret, line 1: After fixing --no-default-features, the CI was failing on errors and it was a huge pain so I fixed it by moving out the wasm code to a separate crate as has slowly been done in other repos.

crowlKats (approved):

PR #413, opened 2023-12-06
fix: ambient namespaces should have members as exports

dsherret:


tests/specs/DeclareNamespace.txt
```diff
@@ -1,11 +1,11 @@
 # mod.ts
 /** Namespace JSdoc */
 declare namespace RootNs {
-    declare const a = "a";
+    const a = "a";
```
dsherret, line 4: TypeScript errors for `declare` within a `declare namespace`.

Issue #427, opened 2023-12-15
html output handle import of namespace and element usage


crowlKats:
`a.ts`
```ts
import { Foo } from "./b.ts";
export function f(): Foo.Data {}
```

`b.ts`
```ts
export namespace Foo {
  export interface Data {}
}
```

Issue #426, opened 2023-12-15
fix handling of aliases imports and exports


crowlKats:
### Import alias case:
`a.ts`
```ts
import { Data as Foo } from "./b.ts";
export interface Data {}
export function f(): Foo {}
```

`b.ts`
```ts
export interface Data {}
```

### Export alias case:
`a.ts`
```ts
import { Data } from "./_util.ts";
export function f(): Data {
```

`b.ts`
```ts
export { type Data as Foo } from "./_util.ts";
```

`_util.ts`
```ts
export interface Data {}
```

cc @lucacasonato

PR #428, opened 2023-12-15
feat: proper import and exports handling

crowlKats:
Before, we had the `AllSymbols` struct, however having all symbols is wrong, and imports and exports need to be dealt with depending on the current file. This also adds a new option to specify how to resolve imports that are external to the current generation.


src/html/util.rs
```diff
@@ -83,39 +86,35 @@ impl NamespacedGlobalSymbols {
...
 
+pub type ImportHrefResolver = Rc<dyn Fn(&[String], &String) -> Option<String>>;
+
 #[derive(Clone)]
 pub struct RenderContext<'ctx> {
-  tt: Rc<TinyTemplate<'ctx>>,
-  all_symbols: NamespacedSymbols,
+  pub(crate) ctx: &'ctx GenerateCtx<'ctx>,
+  current_exports: NamespacedSymbols,
+  current_imports: HashMap<String, String>,
```
crowlKats, line 95: ideally it would re-use `GenerateCtx::global_symbols`, however first #427 needs to be investigated. Potentially easy to do, but perhaps conflicting/simplified by #426

bartlomieju (approved):
Looks good, would be great to add some unit tests are these util functions

PR #441, opened 2023-12-25
refactor(html): move more to templates and contexts

crowlKats:
Depends on #440
Only thing that hasnt been changed yet is namespaces. Will keep that for a follow-up.


bartlomieju (approved):
LGTM, this looks better

PR #454, opened 2024-01-09
fix: add `.ddoc` prefix to tailwind classes output

crowlKats:
`ddoc` for a shorter `deno-doc` to not overly bloat up the tailwind output


lucacasonato:
Even just `dd`?

crowlKats:
that might work, but `dd` is a html element, so maybe just to be clear enough that its something not related to that

lucacasonato:
`.d`? :D

crowlKats:
🤣 a bit too abstract imo. but dont have a strong opinion with any of these, just as long as they are understandable

lucacasonato:
Let's land as is then.

lucacasonato (approved):

PR #457, opened 2024-01-11
fix: actually support serde::Deserialize on DocNode

lucacasonato:
Previously wasn't working, because all the nodes marked as `skip_serializing_if` were not also marked as `default`.

### DIFF
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -2,6 +2,7 @@
 
 use std::path::PathBuf;
 
+use deno_doc::DocNode;
 use deno_graph::source::Source;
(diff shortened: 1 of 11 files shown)


tests/integration_test.rs
```diff
@@ -77,5 +78,9 @@ async fn test_doc_specs() {
       "Should be same for {}",
       test_file_path.display()
     );
+
+    // Check that the JSON output is round-trippable.
+    let _parsed_json_output: Vec<DocNode> =
```
lucacasonato, line 83: This test fails on `main`.

crowlKats (approved):

PR #459, opened 2024-01-11
refactor: more targeted escapes

crowlKats:


src/html/types.rs
```diff
@@ -61,10 +61,13 @@ pub(crate) fn render_type_def(
       let name = if let Some(href) = href {
         format!(
           r#"<a href="{href}" class="link">{}</a>"#,
```
lucacasonato, line 63: `href` is safe because we generate it?

lucacasonato (approved):

src/html/types.rs
crowlKats: yes, though we can escpae it for good measure

PR #460, opened 2024-01-12
refactor: use tree-sitter instead syntect

lucacasonato:
tree-sitter is a much nicer parser, and pulls in much fewer deps.

Also I've done some minor cleanups on the generated HTML, and a refactor how the CSS is generated to further shrink it (passing it through lightningcss).


Cargo.toml
```diff
@@ -40,7 +40,16 @@ termcolor = "1.1.2"
...
-syntect = { version = "5.1.0", optional = true, default-features = false, features = [ "parsing", "default-syntaxes", "default-themes", "html", "dump-load", "regex-onig" ] }
+tree-sitter-highlight = { version = "0.20.1", optional = true }
+
+tree-sitter-javascript = { version = "0.20.1", optional = true }
+tree-sitter-typescript = { version = "0.20.3", optional = true }
+tree-sitter-json = { version = "0.20.1", optional = true }
+tree-sitter-regex = { version = "0.20.0", optional = true }
+tree-sitter-css = { version = "0.20.0", optional = true }
+tree-sitter-toml = { version = "0.20.0", optional = true }
+tree-sitter-md = { version = "0.1.7", optional = true }
```
crowlKats, line 51: seems inferior in terms of supported languages compared to syntect https://github.com/sublimehq/Packages/tree/fa6b8629c95041bf262d4c1dab95c456a0530122 (though i will admit we dont need all these).
does tree-sitter-json  support jsonc?

Cargo.toml
```diff
@@ -50,9 +59,17 @@ tokio = { version = "1.25.0", features = ["full"] }
...
-html = ["html-escape", "comrak", "handlebars", "syntect"]
+html = ["html-escape", "comrak", "handlebars", "tree-sitter-highlight"]
+tree-sitter-all = ["tree-sitter-javascript", "tree-sitter-typescript", "tree-sitter-json", "tree-sitter-regex", "tree-sitter-css", "tree-sitter-toml", "tree-sitter-md"]
+tree-sitter-javascript = ["dep:tree-sitter-javascript"]
+tree-sitter-typescript = ["dep:tree-sitter-typescript"]
+tree-sitter-json = ["dep:tree-sitter-json"]
+tree-sitter-regex = ["dep:tree-sitter-regex"]
+tree-sitter-css = ["dep:tree-sitter-css"]
+tree-sitter-toml = ["dep:tree-sitter-toml"]
+tree-sitter-md = ["dep:tree-sitter-md"]
```
crowlKats, line 72: seems unecessary to have

crowlKats (commented):
how is filesize & perfomance for the example binary compared? (for the latter, test with `deno types --unstable`)

Cargo.toml
lucacasonato: Just always enable them all? Ok.

lucacasonato:
@crowlKats It has much more accurate output than any regexp based syntax highlighter, because it uses a "proper" parser rather than a regexp matcher. It can generally offer the same accuracy as semantic syntax highlighting (which usually requires a pass through TSC or similar LSP server) but at or exceeding the performance of regexp based syntax highlighter.

tree-sitter is also the syntax highlighter used by GitHub to highlight code for most languages.

Having this in the binary would also allow us to implement go-to-definition for JS / TS files in the registry source view at some point.

|  | Before | After |
| - | - | - |
| ddoc size | 11M | 13M |
| ddoc --html types.d.ts | 1.985 s ±  0.036 s | 1.872 s ±  0.046 s |
| generated_docs size | 69M | 71MB |

The generated_docs size difference mainly boils down to the examples now having more accurate highlighting, for example the idents in a member expression are now highlighted correctly:

Before
image

After:
image

Cargo.toml
lucacasonato: Yes: 
image

lucacasonato: The main one that is missing right now that I'd like to have is HTML (this is waiting on a crates.io release: https://github.com/tree-sitter/tree-sitter-html/issues/68).

I couldn't think of other languages that may be useful. But you can find the full list here: https://tree-sitter.github.io/tree-sitter/#parsers

crowlKats: probably should have wgsl, and rust (and WAT in the future) for wasm modules

Cargo.toml
crowlKats: yea, we dont need to conditionally have them in any of our scenarios

crowlKats (approved):

PR #469, opened 2024-01-18
fix: include the byte index in `Location`

lucacasonato:


tests/html_test.rs
```diff
@@ -352,7 +352,7 @@ async fn symbol_search() {
     .join("symbol_search.json");
 
   // uncomment to regenerate symbol_search.json
-  //std::fs::write(&symbol_search_json_path, &file_json);
+  // std::fs::write(&symbol_search_json_path, &file_json);
```
dsherret, line 355: Side note: We should maybe make this use the `UPDATE=1` env var

dsherret (approved):
LGTM

PR #473, opened 2024-01-25
fix(html): better function handling and ignore private symbols in sidepanels & all symbols

crowlKats:


src/html/mod.rs
```diff
@@ -332,6 +332,8 @@ pub fn generate(
       })
       .collect::<Vec<DocNodeWithContext>>();
 
+    dbg!(&all_doc_nodes);
```
lucacasonato, line 335:
```suggestion
```

lucacasonato (approved):

PR #490, opened 2024-02-15
refactor: upgrade to deno_graph 0.66

dsherret:


js/mod.ts
```diff
@@ -2,16 +2,16 @@
 
 import { instantiate } from "./deno_doc_wasm.generated.js";
 import type { DocNode } from "./types.d.ts";
-import { load as defaultLoad } from "https://deno.land/x/deno_graph@0.64.1/loader.ts";
+import { createCache } from "https://deno.land/x/deno_cache@0.7.1/mod.ts";
```
dsherret, line 5: This now uses the cache from deno_cache, which will properly verify JSR checksums. I'm not sure this works on deploy though.

.github/workflows/ci.yml
dsherret, line 1: It uses the deno_cache loader by default which requires more permissions.

crowlKats (approved):
LGTM

PR #492, opened 2024-02-16
chore: update deno_graph 0.67.0

marvinhagemeister:
Looks like cargo formatted the file a little differently. Only updated `deno_graph`


dsherret (approved):
LGTM, but I think prettier auto-formatted the Cargo.toml file.

crowlKats:
i think thats fine: we have no formatting on the Cargo.toml from begin with, which has kinda annoying me over time

PR #494, opened 2024-02-19
fix: images in markdown should be rendered with inline-block

lucacasonato:


src/html/templates/styles.css
```diff
@@ -426,6 +426,10 @@ a {
   text-align: center;
 }
 
+.markdown img {
+  display: inline-block;
+}
+
```
crowlKats, line 432: please nest it in the .mardown class declaration

crowlKats:
not sure why ci didnt error out, since the `styles.gen.css` wasnt updated

crowlKats (approved):

PR #503, opened 2024-02-26
feat: show symbol list in non-main module docs

crowlKats:


src/html/jsdoc.rs
```diff
@@ -334,7 +334,9 @@ impl ModuleDocCtx {
       None
     };
 
-    let (deprecated, examples, docs) = if let Some(node) = module_doc_nodes
+    let mut sections = vec![];
```
bartlomieju, line 337: Maybe pre-allocate to some value?

src/html/templates/section.hbs
```diff
@@ -1,5 +1,7 @@
 <div>
   <h2 class="text-xl leading-6 font-semibold py-1 mb-4">{{title}}</h2>
+  {{print "foo"}}
```
bartlomieju, line 3: This seems like a leftover?

src/html/templates/namespace_section.hbs
```diff
@@ -1,4 +1,7 @@
 <div class="grid grid-cols-1 lg:grid-cols-3 gap-8">
+  {{print "bar"}}
```
bartlomieju, line 2: Ditto

bartlomieju (approved):

PR #508, opened 2024-03-01
fix: properly handle namespaces in all symbols view

crowlKats:


src/html/search.rs
```diff
@@ -33,10 +34,10 @@ fn doc_node_into_search_index_nodes_inner(
       .collect::<Vec<_>>(),
   );
 
-  let name = if ns_qualifiers.is_empty() {
+  let name = if doc_nodes[0].ns_qualifiers.is_empty() {
```
bartlomieju, line 37: Do a check that there's at least one element in `doc_nodes`?

bartlomieju (approved):

PR #510, opened 2024-03-05
feat: expose default export names

crowlKats:


src/function.rs
```diff
@@ -17,6 +17,8 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct FunctionDef {
+  #[serde(skip_serializing_if = "Option::is_none", default)]
+  pub default_name: Option<String>,
```
dsherret, line 21: Maybe we should just store this as something not default export specific (ex. `def_name`, `original_name`, or just `name`)? For exmaple, when a doc node gets renamed we could copy the original name into this property. Keeping the original name might be useful in the future in named export scenarios for some reason:

```
function myFunction() {}

export { myFunction as myNewFunctionName };
```

crowlKats: renamed to `def_name`

src/class.rs
```diff
@@ -243,6 +243,8 @@ impl Display for ClassMethodDef {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct ClassDef {
+  #[serde(skip_serializing_if = "Option::is_none", default)]
+  pub def_name: Option<String>,
```
dsherret, line 247: Maybe we should add a comment about when this is set.

dsherret (approved):
LGTM, though the html code seems lacking in tests.

PR #511, opened 2024-03-06
fix(html): various styling & logic fixes

crowlKats:
Todo:
- [x] code blocks copy button moves with scroll
- [x] code blocks copy button copy content needs to be escaped better


src/html/jsdoc.rs
```diff
@@ -154,9 +154,28 @@ pub fn markdown_to_html(
     let mut ammonia_builder = ammonia::Builder::default();
 
     ammonia_builder
-      .add_tags(["video"])
+      .add_tags(["video", "button", "svg", "path"])
       .add_generic_attributes(["id"])
+      .add_tag_attributes("button", ["onclick"])
```
lucacasonato, line 159: Very bad :sweat_smile: 

Maybe add a `data-copy="thing to copy"` attribute to the button, and then register a global click handler on the document and when invoked, check whether the target has a `data-copy`, and if it does, `e.preventDefault()` and copy?

We can use that for all the copy buttons then.

src/html/jsdoc.rs
```diff
@@ -193,8 +193,18 @@ pub fn markdown_to_html(
...
+    html = ammonia_builder.clean(&html).to_string();
+  }
+
+  html.push_str(
+    r#"<script>
+for (const button of document.getElementsByTagName("button")) {
+  if (button.dataset["copy"]) {
+    button.onclick = () => navigator?.clipboard?.writeText(button.dataset["copy"]);
+  }
+}
```
roj1512, line 205: If I’m not wrong, this is what Luca exactly meant:

```
document.addEventListener("click", (e) => {
  if (e.target instanceof HTMButtonElement && e.target.dataset.copy) {
    return navigator?.clipboard?.writeText(e.target.dataset.copy);
  }
});
```

- It doesn’t override click listeners added by the user, unlike doing `.onclick = ...`.
- It doesn’t add a lot of click listeners.

Sorry for the interference.

lucacasonato: Yes that is what I meant.

crowlKats: the reason i did the `onclick` is because this script will be present for each markdown thats rendered

lucacasonato: Is that more than one per page?

crowlKats: yes. ie each example, each field of a class, each namespace member

crowlKats: however the markdown rendering is the top-most common element. 
another solution would be having a separate template for this script that is added to pages, and JSR then needs to add that to the return payload as well

lucacasonato: Yeah I think that'd be better.

lucacasonato (approved):
LGTM!

Issue #489, opened 2024-02-12
Support for `@see` tag


Im-Beast:
Reference: https://jsdoc.app/tags-see

crowlKats:
Closing as this was finished in #627

PR #514, opened 2024-03-06
feat: `@see` tag

crowlKats:
Towards #489. Does not include html gen handling.

This also slightly changes the semantics of `@category`, `@example` & `@tags`, as they allowed for an empty doc content, which makes no sense for these tags.

### DIFF
--- a/src/js_doc.rs
+++ b/src/js_doc.rs
@@ -613,7 +613,7 @@ if (true) {
       .unwrap(),
       json!({
         "tags": [{
-          "kind":"tags",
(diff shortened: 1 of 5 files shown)


src/js_doc.rs
```diff
@@ -619,6 +618,15 @@ if (true) {
         }]
       })
     );
+    assert_eq!(
```
bartlomieju, line 621: Can you add a test that has several of these tags in a single string?

bartlomieju (approved):

PR #515, opened 2024-03-06
feat: infer function parameter assignment type

### LINKED ISSUE #481: Missing type for arguments if parameter has default value
Given the below code the parameter `from` has a default value of `0` making it an optional number type. However this information is not communicated in the docs generated on `deno.land/x/` or `jsr.io`

```ts
export function* range(from = 0, to?: number): Generator<number> {
    if (to == undefined) {
        to = from
        from = 0
    }
    for (let i = from; i < to; ++i)
        yield i
}
```
Screenshot 2024-02-08 at 18-19-13 range _mod ts iterstar@v1 2 1 Deno
Screenshot 2024-02-08 at 18-19-22 range - @doctor_iterstar - JSR

crowlKats:
Towards #481


src/params.rs
```diff
@@ -236,30 +248,36 @@ fn array_pat_to_param_def(
 }
 
 pub fn assign_pat_to_param_def(
-  parsed_source: Option<&ParsedSource>,
+  parsed_source: &ParsedSource,
   assign_pat: &deno_ast::swc::ast::AssignPat,
 ) -> ParamDef {
+  let infer_ts_type = crate::ts_type::infer_ts_type_from_expr(
```
crowlKats, line 254: all of this just because this function takes a ParsedSource.

src/ts_type.rs
```diff
@@ -25,21 +25,23 @@ use std::fmt::Display;
 use std::fmt::Formatter;
 use std::fmt::Result as FmtResult;
 
-impl From<&TsLitType> for TsTypeDef {
-  fn from(other: &TsLitType) -> TsTypeDef {
+impl From<(&TsLitType, &ParsedSource)> for TsTypeDef {
```
dsherret, line 28: This code is extremely hard to follow. We probably shouldn't be using a `From` implementation on a tuple.

lucacasonato (commented):
this is great!

tests/specs/InferFunctionAssignment.txt
```diff
@@ -0,0 +1,59 @@
+# mod.ts
+export function foo(n = 0) {}
```
dsherret, line 2: Maybe add a test here for `n: string = 0` just to see what it does. It should be `string`, right?

dsherret (approved):
LGTM! Much better without the tuple From implementations

dsherret (approved):
LGTM!

PR #521, opened 2024-03-08
feat(html): display default value of function params

### LINKED ISSUE #481: Missing type for arguments if parameter has default value

crowlKats:
Follow-up to #515.
Closes #481

### DIFF
--- a/src/html/symbols/function.rs
+++ b/src/html/symbols/function.rs
@@ -155,18 +155,29 @@ fn render_single_function(
 
   // TODO: tags
 
-  let param_docs = doc_node
-    .js_doc
-    .tags
-    .iter()
-    .filter_map(|tag| {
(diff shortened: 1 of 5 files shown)


dsherret (approved):
LGTM!

PR #527, opened 2024-03-13
Removed text-left as it clashes with the generated align attribute

furkankly:
`text-align` takes precedence over `align` attribute which is generated by [`markdown_to_html`](https://github.com/denoland/deno_doc/blob/91da46484abf07552c2e993862077b476297b6c9/src/html/jsdoc.rs#L232). Setting this property to `initial !important` in the consumer side didn't work and my guess is that it's the HTML5 spec that doesn't have proper support for `align` attribute anymore. 

Since we are setting it to the default here, removing this should not disrupt anything **unless someone relies on this css for the alignment of nested `p` tags** as its an inherited css prop.


crowlKats (approved):
LGTM, thanks

PR #530, opened 2024-03-18
fix: don't remove all stars from the docs

sigmaSd:
fix https://github.com/denoland/deno_doc/issues/523


sigmaSd:
I noticed that some correct stars still gets stripped like `**Example**`, not sure whats the best way to handle these

crowlKats:
@sigmaSd can you add some tests?

sigmaSd:
I added some tests

src/util/swc.rs
```diff
@@ -124,3 +126,64 @@ pub fn module_export_name_value(
...
+ * # Program
+ * **Example:**
+ *
+ * example1
+ */
+"#
+      ),
+      r#"/**
+# Program
+**Example:**
```
crowlKats, line 183: this doesnt seem to be getting stripped, unlike what you mentioned in your previous comment

sigmaSd: Yes what I had is 
```
/**
** Example** // no starting star 
*/
```

But then I guess maybe we can consider this a user error to not use the star at the start

crowlKats (approved):
LGTM

PR #532, opened 2024-03-26
refactor: use Arc for namespace elements

crowlKats:
required by JSR


lucacasonato (commented):
Why?

crowlKats:
@lucacasonato in #531, we switched namesypace elements to be `Rc`, but I didnt test it on JSR. JSR without this switch to `Arc` errors out because it needs to `DocNode`s to be `Send`.

lucacasonato (approved):

PR #540, opened 2024-04-04
fix(html): fixed horizontal overflow in certain templates

furkankly:
Before:
before
After:
after

I've used `break-words` when possible, `break-all` when the former didn't take effect.


crowlKats (approved):
LGTM

PR #553, opened 2024-04-22
refactor(html): move more tailwind usage to defined classes

crowlKats:


src/html/templates/styles.css
```diff
@@ -24,65 +24,262 @@ summary::-webkit-details-marker {
...
+          }
+        }
+      }
+    }
+  }
+  
+  pre.highlight {
+    @apply border border-gray-300 max-md:border-x-0;
+  }
 }
```
josh-collinsworth, line 245: Question: is this selector nesting compiled away? 

If so, disregard this comment. But otherwise, nesting is fairly well supported, but still new enough it may be risky to use, especially without the `&` syntax. https://caniuse.com/?search=nesting

josh-collinsworth (approved):
Other than one small concern (which may be for nothing), looks all right to me.

src/html/templates/styles.css
crowlKats: yep, its compiled away. I know tailwind supports nesting, but as also pass the stylesheet through lightningcss

PR #558, opened 2024-04-24
fix(publish): update lockfile

crowlKats:


dsherret (approved):
LGTM temporarily, but we should clean up all these duplicate deps.

PR #561, opened 2024-04-25
feat: handle JSDoc for ts type literals

### LINKED ISSUE #302: Type Literal Properties and Methods should support JSDoc
Like classes and interfaces, type literal methods and properties should support having JSDoc.

crowlKats:
Closes #302

### DIFF
--- a/src/ts_type.rs
+++ b/src/ts_type.rs
@@ -13,10 +13,13 @@ use crate::params::ts_fn_param_to_param_def;
 use crate::ts_type_param::maybe_type_param_decl_to_type_param_defs;
 use crate::ts_type_param::TsTypeParamDef;
 use crate::util::swc::is_false;
+use crate::util::swc::js_doc_for_range;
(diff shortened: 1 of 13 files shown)


src/ts_type.rs
```diff
@@ -337,171 +340,206 @@ impl TsTypeDef {
 
       match &type_element {
         TsMethodSignature(ts_method_sig) => {
-          let mut params = vec![];
+          if let Some(prop_js_doc) =
+            js_doc_for_range(parsed_source, &ts_method_sig.range())
+          {
+            let mut params = vec![];
```
dsherret, line 346: Looks like this could be pre-allocated? `let mut params = Vec::with_capacity(ts_method_sig.params.len());`?

dsherret (approved):
LGTM. Just needs some pre-allocates or .iter() w/ .map() and .collect() in a few places

PR #562, opened 2024-04-25
feat: support Location for ts literals, render ts literals as interfaces

crowlKats:
To reduce complexity, `LiteralMethodDef` and related are stripped the `Literal` prefix, and are used for interfaces as well as they now have identical structure. IndexSignature get the same treatment, but additionally also applies to classes.


dsherret (approved):
LGTM

PR #568, opened 2024-04-29
feat: separate constructors from methods for type literals and interfaces

crowlKats:
having constructors be methods called `new` makes it difficult for the html generator to render in a proper manner. example of problem when being a method:
image

one option would be to instead separate methods that called `new` in the html, but doesnt seem desirable, especially as that wouldnt separate between actual method called `new` and constructors.

### DIFF
--- a/tests/specs/interface_construct.txt
+++ b/tests/specs/interface_construct.txt
@@ -11,26 +11,11 @@ error[missing-jsdoc]: exported symbol is missing JSDoc documentation
   | ^
 
 
-error[missing-jsdoc]: exported symbol is missing JSDoc documentation
- --> /mod.ts:2:3
-  | 
-2 |   new(name: string);
-  |   ^
-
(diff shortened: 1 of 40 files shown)


dsherret:
Can you update the PR description with the reason for the change? It breaks the json output I think?

tests/specs/interface_construct.txt
```diff
@@ -11,26 +11,11 @@ error[missing-jsdoc]: exported symbol is missing JSDoc documentation
...
-  |   ^
-
-
-error[missing-return-type]: exported function is missing an explicit return type annotation
- --> /mod.ts:2:3
-  | 
-2 |   new(name: string);
-  |   ^
-
-
```
dsherret, line 27: Seems broken (and below)?

dsherret (approved):
LGTM

PR #573, opened 2024-05-08
test: add snapshot testing of html output

kt3k:
I'm not completely sure this is a good idea, but this PR adds snapshot testing of html output of `html_doc_files` and `html_doc_files_rewrite` test cases using [insta](https://github.com/mitsuhiko/insta) snapshot testing tool.

By these assertions, we can detect any unexpected changes as snapshot diffs in the future html renderer changes.


crowlKats (approved):
LGTM, great to have!

PR #576, opened 2024-05-15
fix: private reference diagnostic was displaying incorrect information

dsherret:
This was also causing a panic in the code.

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -33,17 +33,21 @@ use std::borrow::Cow;
 use std::collections::HashSet;
 use std::rc::Rc;
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub enum DocDiagnosticKind {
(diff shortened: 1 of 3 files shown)


src/diagnostics.rs
```diff
@@ -130,23 +133,23 @@ impl Diagnostic for DocDiagnostic {
...
-          range: DiagnosticSourceRange {
-            start: DiagnosticSourcePos::ByteIndex(
-              reference_location.byte_index,
-            ),
-            end: DiagnosticSourcePos::ByteIndex(
-              reference_location.byte_index + 1,
-            ),
+      DocDiagnosticKind::PrivateTypeRef(diagnostic) => {
+        Some(DiagnosticSnippet {
+          source: Cow::Borrowed(&diagnostic.reference_text_info),
```
dsherret, line 138: Fix is here. It should be using the referenced module's text info and not the original module's text info

crowlKats (approved):
LGTM

PR #581, opened 2024-05-20
feat(html): render private symbols, but dont list them

crowlKats:
This adds the capability to render symbols that arent exported but are referenced by symbols that are public, however the private symbols are not visible on overview pages, and are only reachable via links from other symbols. Additionally, same behaviour is achieved via a jsdoc `@internal` tag for symbols that are exported. 

Also removes the insta test for styles.css, since its not necessary since its a generated file that is clearly visible in PR diffs and not humanly-auditable eitherway


tests/snapshots/html_test__html_doc_files_rewrite-16.snap
```diff
@@ -231,7 +231,7 @@ Type Aliases</h2></div><div class="namespaceSection"><div class="namespaceItem"
 <span class="hover:bg-Interface/15">Hello</span>
               </a>
             </li></ul><a class="flex items-center gap-0.5" href="..&#x2F;.&#x2F;all_symbols.html">
-            <span class="leading-none">view all 6 symbols</span><svg
+            <span class="leading-none">view all 8 symbols</span><svg
```
kt3k, line 234: All symbols page only shows 6 symbols (because we intentionally don't list `A` and `B`). Should this number exclude private and internal symbols?

kt3k (commented):
> Also removes the insta test for styles.css, since its not necessary since its a generated file that is clearly visible in PR diffs and not humanly-auditable eitherway

This makes sense to me.

> This adds the capability to render symbols that arent exported but are referenced by symbols that are public, however the private symbols are not visible on overview pages, and are only reachable via links from other symbols.

Confirmed the doc for `A` class exist and linked correctly from `Foo` document, but it's not listed in overview page/all symbols page.

Left one question about the number of symbols

tests/snapshots/html_test__html_doc_files_rewrite-16.snap
crowlKats: hm, you are right, will update that

kt3k (approved):
LGTM. Nice work!

PR #582, opened 2024-05-21
fix: use "module" identifier for default export item, when the symbol name is not available

kt3k:
Currently we try to show the symbol name using the expression:

```rust
file
  .display_name()
  .replace('-', "_")
  .replace(|c: char| c.is_ascii_alphanumeric(), "")
```
If I invert the last condition to `!c.is_ascii_alphanumeric()`, then it returns something like `ats` for the file `a.ts` (look like trying to guess the name from the last path component) for multi source scenario, but still returns empty string for single file scenario.

I'd suggest we always use `module` identifier for this, which seems less surprising than what the current algorithm tries to do.

The case reported in https://github.com/jsr-io/jsr/issues/459 now becomes like:
Screenshot 2024-05-21 at 20 59 59

closes https://github.com/jsr-io/jsr/issues/459


crowlKats (approved):
I think this is an acceptable approach, though not ideal. lets land this for now, and perhaps in the future we can re-investigate this

kt3k:
BTW this doesn't make all default identifiers "module". In the cases like the below, the usage becomes `import Foobar from ...` (keeping the existing behavior):

```ts
export default function Foobar() {}
```

This only happens when the symbol name is not available.

PR #586, opened 2024-05-24
feat(html): jsdoc module linking

crowlKats:
allows linking to other modules and their symbols. syntax is:
- `{@link [somemodule]}`
-  `{@link [somemodule].somesymbol}`

Ref: https://jsdoc.app/tags-module


src/html/jsdoc.rs
```diff
@@ -656,5 +784,26 @@ mod test {
...
+        parse_links("foo {@link bar} bar", &render_ctx),
+        "foo [bar](../../.././/a.ts/~/bar.html) bar"
+      );
+      assert_eq!(
+        parse_links("foo {@linkcode bar} bar", &render_ctx),
+        "foo [`bar`](../../.././/a.ts/~/bar.html) bar"
+      );
+
+      assert_eq!(
+        parse_links("foo {@link module:b.ts.baz} bar", &render_ctx),
```
lucacasonato, line 800: This is concerning. It is not immediately clear whether this means `baz` export from `b.ts`, or `baz.ts` export from `b`, or the entire `b.ts.baz` module. Ie you can not construct the link without knowing the context of which files are around you.

lucacasonato (commented):
There is no test for linking to a module doc.

I propose alternate syntax:

```
{@link [../foo.ts]} # links to module doc of ../foo.ts
{@link [../foo.ts].bar} # links to bar export from ../foo.ts
```

This diverges from jsdoc, but I think that's fine. The jsdoc syntax for this is very bad.

lucacasonato (approved):
Great, LGTM!

Do we have a docs page on `deno doc` where we can document this?

PR #596, opened 2024-05-29
fix(html): don't dedupe nodes for kind partition

crowlKats:


kt3k:
I tried this patch with https://github.com/denoland/deno_std/blob/main/uuid/v1.ts (which has half deprecated API), but I could see the difference in the ddoc output..

Can you describe the supposed change? Does this affect html output?

crowlKats:
This what the uuid v1 entrypoint page looks like to me now. without the fix its striked through:

Screenshot 2024-05-29 at 21 56 11

kt3k:
Is your testing command like this?

```
cargo run --example ddoc -- --html --output=generated_docs ../deno_std/uuid/v1.ts
```

I still see something like the below:

Screenshot 2024-05-30 at 12 23 02

BTW is your deno_std up to date?

kt3k (approved):
Confirmed the fix in my local env. LGTM 👍

PR #597, opened 2024-06-03
feat(html): treat variable with function as function

crowlKats:
Towards https://github.com/jsr-io/jsr/issues/569


kt3k (approved):
Confirmed this works locally. Nice feature!

I also confirmed `@param` `@returns` works as expected with the example like below:

```ts
/**
 * A function.
 *
 * @param a The input string
 * @param b The input number
 * @returns hello string
 */
export const c = (a: string, b: number): string => "hello";
```

Screenshot 2024-06-05 at 12 17 39

PR #599, opened 2024-06-05
fix(cli): fix stack overflow in Deno doc when namespace exports itself

MujahedSafaa:
This PR fixes: https://github.com/denoland/deno/issues/23711

When executing `deno doc` on a TypeScript file that contains a namespace exporting itself, like in this file example:
https://gist.github.com/crowlKats/1f6ed084bbab0a10267df9ed31564599
```
export namespace test {
    export { test };
}
```
The program exits with the following error:
 **thread 'main' has overflowed its stack
error: process didn't exit successfully: `target\debug\examples\ddoc.exe test3.d.ts` (exit code: 0xc00000fd, STATUS_STACK_OVERFLOW)**

Solution:
I added a check to see if the export is the same as the namespace and, if so, it ignores generating the `DocNode` for it.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -580,6 +580,10 @@ impl<'a> DocParser<'a> {
     let mut handled_symbols = HashSet::new();
 
     for (export_name, export_symbol_id) in symbol.exports() {
+      if export_name == &namespace_name
+      {
+        continue;
+      }
       handled_symbols.insert(UniqueSymbolId::new(
         module_info.module_id(),
         *export_symbol_id,


src/parser.rs
```diff
@@ -580,6 +580,10 @@ impl<'a> DocParser<'a> {
     let mut handled_symbols = HashSet::new();
 
     for (export_name, export_symbol_id) in symbol.exports() {
+      if export_name == &namespace_name
+      {
+        continue;
+      }
```
dsherret, line 586: What happens for the following code?

```ts
export declare namespace test {
    const test: string;
    export { test };
}
```

Or the following:

```ts
export declare namespace test {
    export class test {}
}
```

I think it can't be based on strings, but probably needs to be based on the symbol id or swc id.

src/parser.rs
dsherret, line 1: We should add a test for this in tests/specs (copy and paste another .txt file, modify it, then run `cargo test` to see if it works. You can use `UPDATE=1` env var to update the expected output with the actual output)

dsherret (commented):
Thanks for the PR, but see the comments. Also, would you be able to sign the CLA?

MujahedSafaa:
> Thanks for the PR, but see the comments. Also, would you be able to sign the CLA?

I signed the CLA document

src/parser.rs
MujahedSafaa: I previously attempted to use the symbol ID, but the output varied for certain inputs. For instance:

```
export function describe(name?: string, options?: TestOptions, fn?: SuiteFn): Promise<void>;
export function describe(name?: string, fn?: SuiteFn): Promise<void>;
export function describe(options?: TestOptions, fn?: SuiteFn): Promise<void>;
export function describe(fn?: SuiteFn): Promise<void>;

export namespace describe {
    /**
     * Shorthand for skipping a suite, equivalent to `describe([name], { skip: true }[, fn])`.
     */
    export function skip(name?: string, options?: TestOptions, fn?: SuiteFn): Promise<void>;
    export function skip(name?: string, fn?: SuiteFn): Promise<void>;
    export function skip(options?: TestOptions, fn?: SuiteFn): Promise<void>;
    export function skip(fn?: SuiteFn): Promise<void>;

    /**
     * Shorthand for marking a suite as `TODO`, equivalent to `describe([name], { todo: true }[, fn])`.
     */
    export function todo(name?: string, options?: TestOptions, fn?: SuiteFn): Promise<void>;
    export function todo(name?: string, fn?: SuiteFn): Promise<void>;
    export function todo(options?: TestOptions, fn?: SuiteFn): Promise<void>;
    export function todo(fn?: SuiteFn): Promise<void>;

    /**
     * Shorthand for marking a suite as `only`, equivalent to `describe([name], { only: true }[, fn])`.
     * @since v18.15.0
     */
    export function only(name?: string, options?: TestOptions, fn?: SuiteFn): Promise<void>;
    export function only(name?: string, fn?: SuiteFn): Promise<void>;
    export function only(options?: TestOptions, fn?: SuiteFn): Promise<void>;
    export function only(fn?: SuiteFn): Promise<void>;
}
```
The documentation is only generated for the first line: export function describe(name?: string, options?: TestOptions, fn?: SuiteFn): Promise<void>; The rest are skipped because they share the same symbol ID.

MujahedSafaa: I tested the two examples you mentioned, and encountered a panic on the debug assertion line:
image
After my changes, the function returns Some instead of None:
`self.visibility.get_root_exported_deps(&child_symbol.unique_id()).clone() = Some(
    SymbolDeps(
        {},
    ),
)`
From my understanding I think it is normal in this case to return some and I believe there is no issue with removing the debug assertion code that caused the panic. After removing the debug assertion line and retesting the two examples, everything worked correctly, generating the right documentation as shown in the following images:
 
image
image

MujahedSafaa: Hello @dsherret 
I wanted to follow up on my previous comment. Could you please check my comment above when you get a chance?

PR #605, opened 2024-07-07
fix(docs): fix ignoring generting module documentation when the source code starts with shebang line

MujahedSafaa:
### Summary
This PR fixes the issue where executing `deno doc` on a TypeScript file that starts with a shebang line (`#!/usr/bin/env -S deno run --allow-net --allow-read`) causes the documentation generation to ignore module docs. 

### Problem
Currently, files that include a shebang line skip the module docs, leading to incomplete documentation. This issue occurs because the method `parsed_source.get_leading_comments()` in the `swc.rs` file returns `None` when there is a shebang line. The byte position for the leading comments is not at `program.start()` due to the shebang line, causing the comments to be missed.

### Example
An example of a TypeScript file starting with a shebang line that faces this issue can be seen [here](https://github.com/denoland/deno_std/blob/fa330f69c34f5cd0148af191272561be76983779/http/file_server.ts#L1C1-L1C52).

### Solution
- Added logic to adjust the position of the leading comments when a shebang line is present.
- Refactored the relevant code in `swc.rs` to correctly identify and include leading comments after the shebang line.
- Updated tests to ensure that files both with and without shebang lines are processed correctly.

### Changes
- Modified `swc.rs` to handle the new position of leading comments when a shebang line is present.
- Added tests in `tests/specs` to cover use case with shebang lines.

### Fix issues
- Fixes [#24339](https://github.com/denoland/deno/issues/24339)


MujahedSafaa:
Hello @dsherret @lucacasonato 
Could you please review my PR?

iuioiua:
Please run `cargo fmt`.

MujahedSafaa:
> Please run `cargo fmt`.
Done

crowlKats (approved):
LGTM

PR #608, opened 2024-07-09
fix: do not show [UNSUPPORTED] default in html docs

kt3k:
This PR changes the rendering of ts_type in parameters of functions. Now if the default value is `[UNSUPPORTED]`, then that default value part is not shown in html docs.

BEFORE
Screenshot 2024-07-09 at 14 11 41

AFTER
Screenshot 2024-07-09 at 14 10 33

closes https://github.com/denoland/deno_std/issues/5316


tests/testdata/multiple/a.ts
```diff
@@ -55,3 +55,11 @@ export function qaz(a: number);
 export function qaz(a: string | number) {}
 
 export const c = (): string => "hello";
+
+export function d(
+  foo: number = 1,
+  bar: string = "bar",
+  baz: { hello?: string } = {},
```
iuioiua, line 62: This is another case that displays `[UNSUPPORTED]`. If this PR doesn't fix this case, I wouldn't consider it blocking. It can be fixed in a follow-up PR.
```suggestion
  baz: { hello?: string } = {},
  qux = bar.length,
```

iuioiua (commented):
Great to see this done!

tests/testdata/multiple/a.ts
kt3k: That case seems rendered like the below:

Screenshot 2024-07-09 at 16 21 41

iuioiua: Why are there square brackets?

kt3k: I'm not sure. cc @crowlKats

crowlKats (approved):
LGTM

PR #611, opened 2024-07-10
feat(html): better formatting for code & backport changes from deno-docs

### LINKED ISSUE #592: [HTML] document navigation should reflect the currently selected overload
When viewing a symbol that has many overloads, the document navigation gets a bit... cluttered.

Take a look at the attached screenshot, which has several well-documented overloads for a single function named `titleCase`. **_All_** of the overloads are included in the sidebar, but because only the selected one is visible in the DOM, only a few of the links in the sidebar actually work. And there's no indicator which ones they are, since the same names are repeated for each overload.

The navigation should only include links that are to actively navigable content that's currently visible. Currently this is an accessibility nightmare.

Also, I think the example sections should be consolidated into a single Examples link. It's confusing to have five links that each say "Examples", but only link to a single example each. Just my two cents.

Thanks!

image

### LINKED ISSUE #603: Methods list is challenging to read
Right now a list of methods can be challenging to read with no line breaks and limited formatting
Screenshot 2024-07-03 at 4 55 39 PM

example of some suggested changes (in addition to adding some space between each method)
image

### LINKED ISSUE #604: Design a better label style used in properties list
Screenshot 2024-07-03 at 5 00 17 PM

With the label taking up the left-aligned space it's hard to parse the list of properties or constructions the label is labeling. The label should probably be a bit smaller and sit above or below the property as a superscript, adding more space between properties along the way

maybe something like this
Screenshot 2024-07-03 at 5 09 20 PM

### LINKED ISSUE #606: Change overload selector to inline docs
When you have a large number of overloads it can be hard to understand that they each have different content associated with them and that the overloads themselves are selectable. I propose we drop the content control from this and make each overload an h3 level header with the content associated with it directly underneath it.

Screenshot 2024-07-08 at 4 29 11 PM

crowlKats:
Closes #604
Closes #606
Closes #603
Closes #592
Closes denoland/deno-docs#521


bartlomieju (approved):
LGTM

PR #612, opened 2024-07-12
fix(html): correctly handle examples in document navigation

crowlKats:
Fixes https://github.com/denoland/deno-docs/issues/573.

Fixes 3 things:
- if first element is not the top-most element in a document navigation, it would break since it would only look at the first element's depth, instead of taking the highest ranking element's depth.
- removes the top-level rendering of "Examples" in the document navigation for functions, since functions have examples in each overload instead of at the generic symbol level.
- adds support for individual example titles being rendered in the document navigation, unlike before where it would only render "Examples" without any sub-elements.


bartlomieju (approved):
LGTM

PR #614, opened 2024-07-15
chore: update tree-sitter, and add xml / dtd

lucacasonato:


marvinhagemeister:
Looks like some call signatures of tree-sitter changed.

crowlKats (approved):

PR #616, opened 2024-07-19
perf(html): initialize ammonia once, faster anchorizer, reduce iter collections & vecs

crowlKats:
Also fixes https://github.com/denoland/deno-docs/issues/591


src/html/comrak_adapters.rs
```diff
@@ -232,11 +232,37 @@ pub struct ToCEntry {
...
 }
 
+#[derive(Default)]
+pub struct Anchorizer {
+  map: HashMap<String, i32>,
+  itoa_buffer: itoa::Buffer,
+}
+
+impl Anchorizer {
+  pub fn anchorize(&mut self, s: &str) -> String {
```
kt3k, line 242: Maybe let's link to comrak's anchorize document to explain what this does. Also maybe better mention we do this for perf improvement?

```suggestion
  /// Returns a String that has been converted into an anchor using the GFM algorithm.
  /// This replaces comrak's implementation to improve the performance.
  /// @see https://docs.rs/comrak/latest/comrak/struct.Anchorizer.html#method.anchorize
  pub fn anchorize(&mut self, s: &str) -> String {
```

kt3k (approved):
Confirmed the category panel is correctly hidden < 1024px, and document navigation keep its contents > 640px, addressing https://github.com/denoland/deno-docs/issues/591

LGTM

Screenshot 2024-07-22 at 20 41 02

PR #623, opened 2024-08-27
 feat(html): extract JSDoc summary/title from jsdoc body 

crowlKats:
currently for namespaces and files we extract the title/summary from the first sentence, and say there is no title/summary if there is no first sentence. 
This PR changes this behaviour to find the first sentence in the entire body of the JSDoc, even if its after a codeblock. This change does not affect examples, only module description and namespace-style symbol summary


kt3k (approved):
I updated the integration/snapshot test case to illustrate better what this PR does

LGTM

PR #627, opened 2024-09-12
feat(html): render @see tags

crowlKats:


kt3k:
It looks great if I put the `@see` tag at the end of the jsdoc.

Screenshot 2024-09-13 at 19 37 45

However if I put the `@see` tag in the middle of the jsdoc, the output looks a bit unexpected to me:

The jsdoc
```js
/**
 * ```ts
 * // This code block is ignored when getting the title of this doc
 * const foobar = new Foobar();
 * ```
 *
 * @see https://example.com
 *
 * Foobar docs
 */
export default class Foobar {
}
```

Screenshot 2024-09-13 at 19 41 54

Looks like the paragraph after `@see` tag is considered as a part of it? Is this an intended behavior?

crowlKats:
> Is this an intended behavior?
@kt3k yes, as jsdoc tags are all greedy and capture everything until the next tag

kt3k (approved):
> @kt3k yes, as jsdoc tags are all greedy and capture everything until the next tag

Then this should be fine.

I also added a `@see` tag usage in test fixture.

LGTM

PR #630, opened 2024-09-19
fix(html): proper tag order

crowlKats:
For https://github.com/jsr-io/jsr/issues/724


marvinhagemeister:
Not sure why the CI isn't running.

kt3k (approved):
Confirmed with the local build that this fixed the order of `readonly` and `optional` tags:

BEFORE
Screenshot 2024-09-24 at 19 02 19


AFTER
Screenshot 2024-09-24 at 19 01 49

I think we can land this without test as this only replaces `HashMap` type with `IndexSet` drop-in replacement type (no logic is changed).

PR #631, opened 2024-09-24
fix: ensure safe text slicing boundaries with multi-byte characters

MujahedSafaa:
This commit resolves a panic in the text_info function, which occurs when attempting to slice a string containing non-ASCII characters.

The fix addresses the issue reported in the Deno repository: https://github.com/denoland/deno/issues/23875

Issue:
The problem arises when running deno doc --lint. The code attempts to extract a portion of text from the source code based on a highlight range [start, end]. However, if the text contains non-ASCII characters, such as Chinese characters, a panic occurs. This happens because the code assumes that characters occupy only one byte (as in ASCII), and it incorrectly increments the range start to a fixed value of "+1" to calculate the range end. For multi-byte non-ASCII characters, this leads to an invalid range for the end value, causing a panic during string slicing.

Solution:
The fix adjusts the logic to calculate the end of the range correctly based on the actual byte size of the character, whether it's ASCII or non-ASCII. Instead of always incrementing the range by +1 (start index + 1), the new approach adds the correct byte size of the character (start index + char size of bytes), ensuring that the range is valid for both single-byte and multi-byte characters.

image

### DIFF
--- a/src/diagnostics.rs
+++ b/src/diagnostics.rs
@@ -110,13 +110,21 @@ impl Diagnostic for DocDiagnostic {
   }
 
   fn snippet(&self) -> Option<DiagnosticSnippet<'_>> {
+    let start_byte_index = self.location.byte_index;
+    let start_char_len = &self.text_info.text()[start_byte_index..]
+      .chars()
+      .next()
+      .map(|ch| ch.len_utf8())
+      .unwrap_or(1);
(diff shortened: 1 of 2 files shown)


dsherret (approved):
LGTM. Thanks! Just waiting on https://github.com/denoland/deno_doc/pull/632

PR #635, opened 2024-09-25
feat(html): show method and properties of classes under classes on symbol overview pages

crowlKats:
Also applies to interfaces, and other symbols with sub-items.
Additionally fixes unclosed tags for the documentation navigation.

Screenshot 2024-09-25 at 17 41 45


marvinhagemeister (approved):
LGTM

PR #636, opened 2024-09-25
feat: support module tag value & use for specifying identifier in usage block

crowlKats:


tests/snapshots/html_test__html_doc_files_rewrite-16.snap
```diff
@@ -110,8 +110,8 @@ Variables</h2></div><div class="namespaceSection"><div id="namespace_default" cl
 </section>
 </main><div class="toc">
     <div><div class="usageContent">
-    <h3>Usage</h3><div class="markdown"><pre class="highlight"><code>import * as mod from "foo";
-</code><button class="context_button" data-copy="import * as mod from &quot;foo&quot;;
+    <h3>Usage</h3><div class="markdown"><pre class="highlight"><code>import * as bar from "foo";
+</code><button class="context_button" data-copy="import * as bar from &quot;foo&quot;;
```
kt3k, line 114: 👍

src/html/usage.rs
```diff
@@ -111,11 +113,32 @@ pub fn usage_to_md(
...
+            .get(file)
+            .and_then(|nodes| {
+              nodes
+                .iter()
+                .find(|node| node.kind() == DocNodeKind::ModuleDoc)
+            })
+            .and_then(|node| {
+              node.js_doc.tags.iter().find_map(|tag| {
+                if let JsDocTag::Module { name } = tag {
+                  name.as_ref().map(|name| name.to_string())
```
kt3k, line 132: I think the name might be causing a syntax error in import statement like `import * as foo-bar from "module";`. Can we avoid it somehow? (Maybe by removing non-alnum character from the name?)

crowlKats: yea was thinking about that; I guess we could allow `[a-zA-Z$_]`, and replace any unknown characters with `_`. more is technically allowed, but i think this is good enough

kt3k (approved):
(Sorry for the delay in review)

The regex replacement makes sense to me. LGTM!

PR #644, opened 2024-10-08
feat: use only tree-sitter

crowlKats:


kt3k (approved):
Looks nice and reasonable simplification to me.

LGTM

PR #652, opened 2024-10-30
feat: add feedback on copy

mbhrznr:
fixes https://github.com/jsr-io/jsr/issues/730.

there hasn't been much of any feedback for the user that copying the code examples was successful.
this pr adds some immediate feedback in the form of a `check` icon, as seen in jsr's `CopyButton.tsx`.

~additionally `write_button` created and appended an empty `code` block, which is now removed.~


src/html/comrak_adapters.rs
```diff
@@ -85,11 +85,11 @@ impl HighlightAdapter {
...
       output,
-      r#"<button class="context_button" data-copy="{}">{}</button>"#,
+      r#"<button class="context_button" data-copy="{}">{}{}</button>"#,
       html_escape::encode_double_quoted_attribute(source),
-      include_str!("./templates/icons/copy.svg")
-    )?;
-    write!(output, "<code>")
+      include_str!("./templates/icons/copy.svg"),
+      include_str!("./templates/icons/check.svg")
+    )
```
crowlKats, line 92: missing opening `code` tag as there was before

mbhrznr: the removal was intentional, as it created an empty `code` block.
should i reintroduce it?

mbhrznr: fair enough! reverted that change, thanks for the hint!

crowlKats (approved):
LGTM, thanks!

PR #655, opened 2024-11-07
feat: dont underscore numbers when formatting module names

c43721:
Fixes https://github.com/jsr-io/jsr/issues/803 but I'd like to know how or if I could better test this. 

I eventually got here by looking at how the identifier for the file was created:
image

image

The identifier replacing all non-characters with underscores seems to be the source so adjusting the regex to allow numbers. 

Though this regex might need to account for a leading number and replace that since you can't have a module identifier prefixed by a number (must be a character)


c43721:
I'm not sure the regex would want to handle something as complex as "allow a-Za-z0-9 but if the string starts with 0-9 then prepend `_`" and I'm not sure that would be a valid export so I don't think that should be a concern here. Any inputs?

crowlKats:
@c43721 yes it should handle the case of a starting number

crowlKats:
@c43721 for test, you can update the module doc name in tests/testdata/multiple/b.ts

c43721:
I'm just going to re-do this under a new branch.

PR #656, opened 2024-11-07
refactor(html): extract and isolate markdown rendering

crowlKats:
- make markdown and highlighting handling pluggable, while still keeping comrak as an option
- completely removes tree-sitter & syntect from the codebase, still pluggable via the comrak path, just not built-in
- make usage composer more ergonomic and isolated
- add an option to inject content into the html head

This is preliminary work for being able to have the html generation aspect available via WASM


kt3k:
Now the usage line in the right navigation area doesn't seem having its own horizontal scroll bar. Instead the entire navigation area scrolls horizontally. This feels a strange UI pattern to me. What do you think?

Screenshot 2024-11-08 at 19 20 11

This scrolls like:

Screenshot 2024-11-08 at 19 20 20

---

Also `{@linkcode ...}` directive now doesn't seem having highlight as before. Is this intended change?

Screenshot 2024-11-08 at 19 24 08

Screenshot 2024-11-08 at 19 24 18

---
Also the code blocks now don't seem having borders and padding. Is this change intentional?

CURRENT
Screenshot 2024-11-08 at 19 25 35

THIS BRANCH
Screenshot 2024-11-08 at 19 25 47

crowlKats:
@kt3k apologies, forgot to load the comrak css file in the example binary, fixed now

kt3k (approved):
Confirmed all points addressed! LGTM!

PR #658, opened 2024-11-08
feat: support html generation via Wasm

### LINKED ISSUE #513: Plans to support HTML in the WASM / JS implementation?
Hey Deno Team,

I'm curious if there's any plans on the horizon to support HTML docs generation from within the WASM/JS implementation. Not sure if this is even possible... so forgive me if I sound like a noob, but being able to generate HTML docs programmatically would be a dream come true. 

My primary need for a feature like this is probably a bit niche. But it's for a custom GPT I'm currently working on, which uses an external API endpoint to provide users with the in-chat capabilities to format users code, write them a README, and _hopefully_ generate a static HTML site for their documentation. 

The API endpoint runs on Deno Deploy, so it goes without saying I don't have access to the CLI features. I've got the formatting worked out with the [`deno_fmt`](https://deno.land/x/deno_fmt) module I published somewhat recently (forgive the shameless plug), but the HTML docs aspect of it has me a bit stuck. Having a feature like this would save me quite a bit of time and energy, so I don't have to re-invent the wheel and write my own HTML codegen from the deno doc JSON.

I assume it would have to be outputted similarly to the `doc()` output, or the `transpile()` output from `deno_emit` (mapping file paths to code), since accessing the local file system would be out of its scope.

Like I said, not sure if it's even possible. My knowledge of Rust is severely limited, but I'm looking to fix that this year 😜.

Anyway, that's enough rambling. Thank you for your time!

crowlKats:
Closes #513


src/html/mod.rs
```diff
@@ -70,6 +72,60 @@ const FUSE_FILENAME: &str = "fuse.js";
 const SEARCH_JS: &str = include_str!("./templates/pages/search.js");
 const SEARCH_FILENAME: &str = "search.js";
 
+#[cfg(target_arch = "wasm32")]
+trait WasmToFilePath {
+  fn to_file_path(&self) -> Result<PathBuf, ()>;
+}
+
+#[cfg(target_arch = "wasm32")]
+impl WasmToFilePath for url::Url {
```
dsherret, line 81: Maybe use deno_path_util crate?

crowlKats: oh wasnt aware it had that functionality built in. updated

dsherret (approved):
LGTM!

PR #659, opened 2024-11-15
feat(wasm/html): add callback to generate the default path resolver in resolvePath

crowlKats:
and includes a few minor adjustments

### DIFF
--- a/src/html/util.rs
+++ b/src/html/util.rs
@@ -232,8 +232,12 @@ impl NamespacedGlobalSymbols {
 pub enum UrlResolveKind<'a> {
   Root,
   AllSymbols,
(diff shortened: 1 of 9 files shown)


src/html/util.rs
```diff
@@ -232,8 +232,12 @@ impl NamespacedGlobalSymbols {
...
   Root,
   AllSymbols,
-  Category(&'a str),
-  File(&'a ShortPath),
+  Category {
+    category: &'a str,
+  },
+  File {
+    file: &'a ShortPath,
+  },
```
crowlKats, line 240: this is done because serde cant handle tuple-style enum variants when using the `tag` attribute

PR #668, opened 2024-11-26
chore: create import map for dependencies

BlackAsLight:
This pull request makes the TypeScript files use an import map instead of them all listing their dependencies and versions individually. Making dependency management easier, and hopefully if https://github.com/denoland/deno_cache_dir/pull/67 gets merged, another pull request to update `@deno/cache-dir` so those warnings mentioned in that pull request go away in `@std`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -22,5 +22,14 @@
       "src/html",
       "tests/testdata"
     ]
+  },
+  "imports": {
+    "@david/dax": "jsr:@david/dax@0.40.1",
+    "@deno/cache-dir": "jsr:@deno/cache-dir@^0.11.1",
+    "@deno/graph": "jsr:@deno/graph@^0.82.3",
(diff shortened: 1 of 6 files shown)


deno.json
```diff
@@ -23,5 +23,14 @@
...
       "tests/testdata"
     ]
+  },
+  "imports": {
+    "@david/dax": "jsr:@david/dax@0.40.1",
+    "@deno/cache-dir": "jsr:@deno/cache-dir@^0.11.1",
+    "@deno/graph": "jsr:@deno/graph@^0.82.3",
+    "@std/assert": "jsr:@std/assert@^0.223.0",
+    "browserslist": "npm:browserslist@4.23.0",
+    "lightningcss": "npm:lightningcss@^1.28.2",
```
kt3k, line 33: Maybe this needs to be `1.26.0` to pass the CI check?

Or, alternatively, we can commit the new css files to avoid css diffs

PR #674, opened 2024-12-05
feat: node references

crowlKats:


chat.ts
dsherret, line 1: Left over?

namespace.ts
dsherret, line 1: Also remove?

src/html/mod.rs
```diff
@@ -408,6 +464,59 @@ impl GenerateCtx {
 
     self.href_resolver.resolve_path(current, target)
   }
+
+  // TODO: don't reference to another node, but redirect to it instead
```
dsherret, line 468: Add your name to the todo?

src/node.rs
```diff
@@ -164,6 +144,9 @@ pub enum DocNodeDef {
     import_def: ImportDef,
   },
   ModuleDoc,
+  Reference {
+    reference_def: ReferenceDef,
+  },
```
dsherret, line 149: 👍

src/parser.rs
```diff
@@ -166,51 +155,244 @@ impl<'a> DocParser<'a> {
...
+    let mut all_locations = doc_nodes_by_url
+      .values()
+      .flatten()
+      .flat_map(|node| {
+        fn walk_rc_nodes(nodes: &[Rc<DocNode>]) -> Vec<Location> {
+          nodes
+            .iter()
+            .flat_map(|node| {
+              if let Some(namespace) = node.namespace_def() {
+                walk_rc_nodes(&namespace.elements)
```
dsherret, line 188: Is it possible for someone to create a circular reference by being stupid and cause an inifinte loop here?

src/parser.rs
```diff
@@ -166,51 +155,244 @@ impl<'a> DocParser<'a> {
...
+    let mut i = 0;
+    while i < nodes.len() {
+      let name = nodes[i].name.to_string();
+
+      match &mut nodes[i].def {
+        DocNodeDef::Namespace { namespace_def } => {
+          let mut namespace_elements = namespace_def
+            .elements
+            .iter()
+            .map(|node| Rc::unwrap_or_clone(node.clone()))
```
dsherret, line 236: `unwrap_or_clone` will always clone? Maybe just do:

```suggestion
            .map(|node| node.as_ref().clone())
```

src/parser.rs
```diff
@@ -166,51 +155,244 @@ impl<'a> DocParser<'a> {
...
+            .collect::<Vec<_>>();
+
+          let mut name_path = name_path.to_vec();
+          name_path.push(name);
```
dsherret, line 240: Nitpick, but this will allocate twice. Maybe set the capacity as name_path.len() + 1 first?

src/parser.rs
```diff
@@ -166,51 +155,244 @@ impl<'a> DocParser<'a> {
...
+    match module {
+      Module::Js(_) | Module::Json(_) | Module::Wasm(_) => {
+        let module_info = self.get_module_info(module.specifier())?;
+        let exports = module_info.exports(&self.root_symbol);
+        let root_name = name_path.remove(0);
+        let (mut export_name, export) = exports
+          .resolved
+          .iter()
+          .find(|(name, _)| &&root_name == name)
+          .unwrap();
```
dsherret, line 297: Maybe let's not unwrap for this one? It could cause issues for bad references?

src/parser.rs
```diff
@@ -166,51 +155,244 @@ impl<'a> DocParser<'a> {
...
+                  let definitions = self
+                    .root_symbol
+                    .go_to_definitions(export.module, export_symbol)
+                    .collect::<Vec<_>>();
+
+                  return self.resolve_dangling_reference(
+                    definitions.first().unwrap().module.specifier(),
+                    reference_def,
+                    // -1 to include the root
+                    name_path[i - 1..].to_vec(),
```
dsherret, line 339: Is it possible for `i` to be `0`?

src/parser.rs
```diff
@@ -166,51 +155,244 @@ impl<'a> DocParser<'a> {
...
+
+              definitions = self
+                .root_symbol
+                .go_to_definitions(export.module, export_symbol)
+                .collect::<Vec<_>>();
+            }
+          }
+        }
+
+        let original_range = &export_symbol.decls().first().unwrap().range;
```
dsherret, line 356: What happens if it refers to multiple declarations? What if those declarations are found in different files? Do we have a test for that?

src/parser.rs
```diff
@@ -166,51 +155,244 @@ impl<'a> DocParser<'a> {
...
+                      return Ok(Some(maybe_docs));
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+
+        unreachable!()
```
dsherret, line 392: Seems risky.

dsherret (commented):
Looks really good! Just have a few questions.

src/parser.rs
crowlKats: changed to debug assertion only and returns none on prod. though this really shouldnt hit

src/parser.rs
crowlKats: i'll add a test

src/parser.rs
crowlKats: how could there be a bad reference though?

src/parser.rs
crowlKats: that should be handled by tests/specs/namespace_reexport_member.txt

src/parser.rs
crowlKats: no since a reference can only be from another file, the first iteration of the loop always passes first for the symbol that is a reference, and only once a reference is followed can this codepath hit, which would happen on second iteration, so its safe. if preferred i can add a fallback that just returns none, but shouldnt happen

src/parser.rs
dsherret: I guess it's not possible.

dsherret (approved):
LGTM. Nice work

KnorpelSenf:
This makes it impossible to generate documentation for files imported via URL.
```
Trace: Error: relative URL without a base
    at __wbg_new_28c511d9baebfa89 (https://jsr.io/@deno/doc/0.184.0/deno_doc_wasm.generated.js:678:19)
```

```ts
// works (version prior to this PR):
await (await import("jsr:@deno/doc@0.161.0")).doc("https://raw.githubusercontent.com/grammyjs/grammY/refs/heads/main/src/mod.ts")
// fails with the above error (version after this PR):
await (await import("jsr:@deno/doc@0.162.0")).doc("https://raw.githubusercontent.com/grammyjs/grammY/refs/heads/main/src/mod.ts")
```

PR #686, opened 2025-01-03
feat(html): allow generating JSON representation of HTML output

crowlKats:


src/html/mod.rs
```diff
@@ -1145,6 +1145,190 @@ pub fn generate(
   Ok(files)
 }
 
+pub fn generate_json(
```
crowlKats, line 1148: this blob has a decent amount of duplicated logic. it is not often-changing, but still. i'll think about how this could be deduplicated in a follow-up

src/html/mod.rs
```diff
@@ -1145,6 +1145,190 @@ pub fn generate(
...
+    let (partitions_for_entrypoint_nodes, uses_categories) =
+      if let Some(entrypoint) = ctx.main_entrypoint.as_ref() {
+        let nodes = ctx.doc_nodes.get(entrypoint).unwrap();
+        let categories = partition::partition_nodes_by_category(
+          &ctx,
+          nodes.iter().map(Cow::Borrowed),
+          ctx.file_mode == FileMode::SingleDts,
+        );
+
+        if categories.len() == 1 && categories.contains_key("Uncategorized") {
```
bartlomieju, line 1164: Consider cleaning this up later to use `static UNCATEGORIZED: &str = "Uncategorized"`

bartlomieju (approved):

PR #693, opened 2025-01-27
fix: show private symbols in dts files

crowlKats:
Also removes the `"default"` header in the all symbols page if its a single d.ts file


src/html/mod.rs
```diff
@@ -803,8 +803,9 @@ impl DocNodeWithContext {
     ns_qualifiers
   }
 
-  pub fn is_internal(&self) -> bool {
-    self.inner.declaration_kind == crate::node::DeclarationKind::Private
+  pub fn is_internal(&self, ctx: &GenerateCtx) -> bool {
+    (self.inner.declaration_kind == crate::node::DeclarationKind::Private
+      && !matches!(ctx.file_mode, FileMode::SingleDts | FileMode::Dts))
```
crowlKats, line 808: this is the main fix

thisisjofrank (approved):
Looks good to me

PR #696, opened 2025-01-29
refactor: remove `DocNodeKind`

crowlKats:
This doesnt affect JSON output, only the rust API.
This is done in favour of just operating of `DocNodeDef` variants instead of having a special type that mirrors all the variants

Also changes `NamespaceDef` elements to be stored as `Arc` instead of `Rc`.

the html gen still has a `DocNodeKindWithDrilldown` (renamed to `DocNodeKind` in this PR for simplicity), which is mainly used for ordering, which should be moved to ordering on `DocNodeDef` and then also add ordering on `DocNode` since we need that in a few spots.
Additionally its also used for grouping of nodes, but that could be cleaned up to use `DocNodeKindCtx` instead.
However not doing that in this PR.

### DIFF
--- a/examples/ddoc/main.rs
+++ b/examples/ddoc/main.rs
@@ -3,11 +3,12 @@
 use clap::App;
 use clap::Arg;
 use deno_doc::find_nodes_by_name_recursively;
+use deno_doc::html::GenerateCtx;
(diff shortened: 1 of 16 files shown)


dsherret (approved):
LGTM

PR #697, opened 2025-02-04
feat(html): introduce dark theme

AugustinMauroy:


crowlKats:
@AugustinMauroy looks really good, only thiing i'd like to see changed before landing is the white border is too much contrast on various elements you can see in this image:
image

AugustinMauroy:
updated highlight border and context button border 😁

crowlKats (approved):

PR #706, opened 2025-03-17
fix: do not stack overflow when namespace exports itself or an ancestor

dsherret:
This just ignores self or ancestor exports for the time being.

Closes https://github.com/denoland/deno_doc/pull/599/


crowlKats (approved):
LGTM

PR #710, opened 2025-04-11
feat: dont underscore numbers when formatting module names

c43721:
Redo of https://github.com/denoland/deno_doc/pull/655

Fixes https://github.com/jsr-io/jsr/issues/803 but I'd like to know how or if I could better test this. 

I eventually got here by looking at how the identifier for the file was created:
image

image

The identifier replacing all non-characters with underscores seems to be the source so adjusting the regex to allow numbers. 

Though this regex might need to account for a leading number and replace that since you can't have a module identifier prefixed by a number (must be a character)

Testing strategy includes running the snapshot tests and asserting both the module name is `bar123` as well as locally inspecting the generated HTML to see `import ... as bar123 from ...`


crowlKats:
@c43721 apologies for the long delay on this. would you be able to sign the new CLA system, so i can merge the PR?

c43721:
@crowlKats You got it. Signed.

crowlKats (approved):
LGTM, thanks!

PR #716, opened 2025-05-13
fix: deno_ast 0.47

dsherret:


Cargo.toml
```diff
@@ -76,3 +76,4 @@ codegen-units = 1
 incremental = true
 lto = true
 opt-level = "s"
+
```
bartlomieju, line 79: On purpose?

bartlomieju (approved):

Cargo.toml
dsherret: Nope. Reverted

PR #718, opened 2025-05-15
fix(html): unified ids and id prefixes

crowlKats:
This adds a more unified and congruent approach for generating IDs. we need this for more predictable IDs so we can do scope and global symbol search on JSR


kt3k (approved):
LGTM 👍

PR #737, opened 2025-08-13
Add missing TypeScript interfaces and properties

kossnocorp:
Hey! I started adopting `deno doc` for date-fns documentation generation. Thank you a lot for the outstanding tool!

After running `deno doc --json` against date-fns and then type checking the resulting JSON, I found that some TypeScript interfaces and properties were missing, so I added them.

I confirmed that the type changes fix all the problems for date-fns by manually updating `node_modules/@deno/doc/types.d.ts`.

---

Here's the script I used to generate and type check the docs JSON: [`buildDocs.sh`](https://github.com/date-fns/docs/blob/main/scripts/buildDocs.sh).

You can see the TypeScript errors I get: [`logs.txt`](https://gist.github.com/kossnocorp/010eab15f91edc9b5dd31eb3603aa387).

Here's the source code I got as the result: [`docs.ts`](https://1drv.ms/t/c/4b4a67abbc59c668/EY6xZhnJKVVLqIqz_RGtib0BL-7e21R-pvyylVQOKpwLyg?e=GcwjWN) (I apologize for a OneDrive link. GitHub Gist was throwing `500` on an attempt to upload ~200K LoC file).

You can also reproduce the problem if you clone the [date-fns repo](https://github.com/date-fns/date-fns) and run:

```sh
git submodule update --recursive --init --remote
pnpm install
cd submodules/docs
DATE_FNS_DIR="../.." ./scripts/buildDocs.sh
```

You will find the `docs.json` at `submodules/docs/tmp/docs.json` and the TypeScript code at `submodules/docs/tmp/docs.ts`.

You can also rerun the `tsc` (after running the `buildDocs.sh` first):

```sh
cd submodules/docs
pnpm tsc --noEmit --skipLibCheck --module NodeNext --moduleResolution NodeNext tmp/docs.ts
```

---

I made missing properties optional to avoid breaking any libraries that depend on it. I also couldn't figure out the proper `constructors` interface, so I made it `unknown[]`.

Furthermore, I also added the `Docs` interface to represent the root data structure of the JSON.


js/types.d.ts
```diff
@@ -1,5 +1,10 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
+export interface Docs {
+  version: 1;
+  nodes: DocNode[];
+}
+
```
crowlKats, line 7:
```suggestion
```

this is not defined in deno_doc, but rather deno cli itself. deno_doc doesnt maintain a version of its json format since its a library that is versioned itself

PR #752, opened 2025-12-25
fix(html): move statics above non-statics

crowlKats:


kt3k (approved):
LGTM when snapshot expectations are updated

PR #753, opened 2025-12-25
feat: jsdoc priority tag

crowlKats:
This PR adds a new jsdoc tag: `@priority`.
It takes a number value, and acts like `z-index` in css does: a bigger number means it has higher priority, 0 same as unset and negative lower priority.

The priority affects in what order symbols are displayed in symbol lists in the HTML renderer, with greater priority showing up first.
This is especially useful for JSR, where there may be some exports you want to prioritize (or "feature").

### DIFF
--- a/src/html/partition.rs
+++ b/src/html/partition.rs
@@ -244,6 +244,34 @@ fn compare_node(
   (!node2_is_deprecated)
     .cmp(&!node1_is_deprecated)
     .then_with(|| {
+      let p1 = node1.js_doc.tags.iter().find_map(|tag| {
+        if let JsDocTag::Priority { priority } = tag {
+          Some(priority)
+        } else {
+          None
+        }
+      });
+      let p2 = node2.js_doc.tags.iter().find_map(|tag| {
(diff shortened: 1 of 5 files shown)


src/html/partition.rs
```diff
@@ -243,6 +243,34 @@ fn compare_node(
...
+        if let JsDocTag::Priority { priority } = tag {
+          Some(priority)
+        } else {
+          None
+        }
+      });
+
+      match (p1, p2) {
+        (Some(p1), Some(p2)) => p1.cmp(p2),
+        (Some(p1), None) if p1 == &0 => Ordering::Equal,
```
kt3k: Looks reasonable to handle None being equal to 0

kt3k (approved):
This is useful for organizing API docs page (especially in JSR) in more reasonable form. I think this is worth trying. LGTM

kt3k:
BTW do we document what tags are supported in `deno doc` or in JSR in somewhere?

crowlKats:
Yes, https://docs.deno.com/runtime/reference/cli/doc/#supported-tags

PR #759, opened 2026-01-27
feat: hide private class members

KnorpelSenf:
As discussed in https://github.com/denoland/deno_doc/pull/757#issuecomment-3802107432.


crowlKats (approved):
LGTM. the current way we handle private is annoying as its quite fragmented and inconsistent, i will completely rework it soon

PR #760, opened 2026-01-30
feat: render jsdocs as markdown with syntax highlighting

dsherret:
This now renders markdown in jsdocs for the terminal output.

Most of this code is generated by claude, with a lot of back and forth.

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -1026,9 +1026,9 @@ fn render_markdown(
             if let Some(p) = saved_prefix {
               self.line_prefix_stack.push(p);
             }
-            self.line_prefix_stack.push(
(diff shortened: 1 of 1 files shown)


tests/specs/indented_with_tabs.txt
```diff
@@ -30,17 +30,21 @@ Defined in file:///mod.ts:8:1
...
   Line 1
-  
+
   Line 2
-  
-  	Indented
+
+  ```
+  Indented
+  ```
```
dsherret: I think this is fine because it's technically a code block?

src/printer.rs
dsherret: In the future, it would be neat to have the ability to render code blocks with syntax highlighting, but this is good enough for now.

PR #769, opened 2026-02-15
fix: various minor fixes

### LINKED ISSUE #434: class expressions are not supported
While this is supported and docs for it generate fine:

```js
class Square extends Rectangle {
	constructor({width}) {
		super({width, height: width});
	}
}
```

This syntax doesn't seem to be supported and the docs generated for it say it is a "variable" and not a class:

```js
const Square = class extends Rectangle {
	constructor({width}) {
		super({width, height: width});
	}
};
```

Note that JSDoc supports both class declarations and expressions: https://jsdoc.app/howto-es2015-classes

### LINKED ISSUE #625: Support for added properties in functions to show as functions
It seems that the properties added are supported but when they are functions they are not treat them like one.

#### Example

The following code:
```ts
/**
 * Test function.
 *
 * @returns {string} - Returns a string with 'hello world'.
 */
export function test(): string {
  return 'hello world';
}

/**
 * Variable inside the test function.
 */
test.innerVariable = 'inner variable';

/**
 * Function inside the test function.
 *
 * @returns {string} - Returns a string with 'inner function'.
 */
test.innerFunction = (): string => 'inner function';
```

Returns the following html:

Screenshot 2024-09-06 at 3 01 01 PM

Both `innerVariable` and `innerFunction` are shown as variables even though the `innerFunction` should be shown as a function.

This is a problem since `innerFunction` is not well documented:

Screenshot 2024-09-06 at 3 04 32 PM

### LINKED ISSUE #626: [json] erroneous parsing of decorators in JSDoc code blocks
When using `deno doc --json`, example code blocks that contain decorators prove to be problematic for the parser. I'm not sure if this problem persists in `--html` or in the pretty-printed format in the CLI, but I would assume so since they all use the same parser under the hood AFAIK.

It stems from the `@` prefix, which has also been a long running problem with the built-in JSDoc parser used by VSCode's intellisense, and a lot of other projects that incorporate JSDoc comment parsing in their logic. The parser gets confused and thinks the decorator is actually a malformed JSDoc tag, leading to unexpected (and usually terribly formatted) results.

Here's an example from the module doc comment in a JSR package of mine. Note the `"unsupported"` tags, particularly the last one. You can see it's parsing tags from inside an example code block.

```jsonc
{
  "version": 1,
  "nodes": [
    // ...
    {
      "name": "",
      "location": {
        "filename": "file:///workspaces/decorators/packages/types/mod.ts",
        "line": 1,
        "col": 0,
        "byteIndex": 0
      },
      "declarationKind": "export",
      "jsDoc": {
        "tags": [
          {
            "kind": "module"
          },
          {
            "kind": "unsupported",
            "value": "@log(\"class\") class Example {\n// ...\n"
          },
          {
            "kind": "unsupported",
            "value": "@log(\"method\") method() { return \"foo\" }\n"
          },
          {
            "kind": "unsupported",
            "value": "@log(\"getter\") get field() { return 42 }\n}\n```"
            // ---- the tell-tale sign its in an example code block  ^^^
          }
        ]
      },
      "kind": "moduleDoc"
    }
  ]
}
```

### LINKED ISSUE #694: Incorrect printing of literal types
**Bit ints**

This code:

```ts
export type Falsy = false | "" | 0 | 0n | null | undefined;
```

incorrectly prints as:

```ts
type Falsy = false | "" | 0 | 0 | null | undefined
```

A big int 0 should print (I think) as `0n`.  TypeScript distinguishes between the types `0` and `0n`.  `0 | 0` will simplify to `0`, while `0 | 0n` will not.

**Strings**

It seems that no escaping is performed on string literal types.  This code:

```ts
export type Newline = "\n";
```

prints (in command-line text output) as:

```
type Newline = "
"
```

### LINKED ISSUE #715: ANSI colors being included in the `tsType` field of some type predicates?
I've been encountering some pretty strange behavior with Deno Doc where generated function doc nodes contain ANSI color sequences like `\u001b[38;5;12m` in their `returnType` fields. I've confirmed this to be occurring in both the WebAssembly build and the Deno CLI version in `--json` mode, as recently as version 0.172.0 / Deno v2.3.1+5044f2f.

```ts
import { doc } from "jsr:@deno/doc@0.172.0";

const specifiers = [
  "jsr:@nick/is@0.2.0-rc.6/both",
  "jsr:@nick/is@0.2.0-rc.6/either",
  "jsr:@nick/is@0.2.0-rc.6/constructor",
];

const result = await doc(specifiers);

console.log(result[specifiers[0]][1].functionDef.returnType);
```

When I run the above code I get the following result. Notice line 7.

```js
{
  repr: "",
  kind: "fnOrConstructor",
  fnOrConstructor: {
    constructor: false,
    tsType: {
      repr: "it is \x1b[0m\x1b[38;5;12mL\x1b[0m & \x1b[0m\x1b[38;5;12mR\x1b[0m",
      kind: "typePredicate",
      typePredicate: {
        asserts: false,
        param: { type: "identifier", name: "it" },
        type: { repr: "", kind: "intersection", intersection: [Array] }
      }
    },
    params: [
      {
        kind: "identifier",
        name: "it",
        optional: false,
        tsType: { repr: "unknown", kind: "keyword", keyword: "unknown" }
      }
    ],
    typeParams: []
  }
}
```

At first I thought it was only occuring on the `isBoth` and `isEither` guards, but I have just confirmed it is happening with other ones as well. The `/constructor` module, third in the `specifiers` array in the above example snippet, generates this doc node structure:

```js
{
  repr: "it is \x1b[0m\x1b[38;5;12mConstructor\x1b[0m<\x1b[0m\x1b[38;5;12mT\x1b[0m>",
  kind: "typePredicate",
  typePredicate: {
    asserts: false,
    param: { type: "identifier", name: "it" },
    type: {
      repr: "Constructor",
      kind: "typeRef",
      typeRef: { typeParams: [ [Object] ], typeName: "Constructor" }
    }
  }
}
```

### LINKED ISSUE #728: `export * as alias from` is not-documentable
The jsdoc seems to be ignored

```ts
/** Re-exports of zod. */
export * as is from "npm:zod"
```

```
error[missing-jsdoc]: exported symbol is missing JSDoc documentation
 --> /workspaces/mod.ts:3:8
  | 
3 | export * as is from "npm:zod"
  | 
```

Note that the workaround of 
```ts
import * as a from "npm:zod"
const b = a as typeof a
export { b }
```

is not practical (in addition to not being elegant) because you loose the types namespace, so for example in zod you cannot do `zod.infer` anymore

### LINKED ISSUE #800: If a library has a single export, show that instead of `import * as _library_name from`
If a library has a simple export, such as https://jsr.io/@nic/local-highlight-registry which only exports `LocalHighlightRegistry`, it would be much more useful if the sidebar showed

```js
import { LocalHighlightRegistry } from "@nic/local-highlight-registry";
```
rather than
```js
import * as _nic_local_highlight_registry from "@nic/local-highlight-registry";
```

By doing so, users can just copy-paste the snipped without having to then modify it, making the "copy" button actually useful :)

This should only be done when there is a single export, since with multiple exports the example would probably become too long and users would likely still have to modify it manually.

### LINKED ISSUE #914: Docs generator considers the hyphen in @param as a bullet point
```
@param {Foo} foo - The foo
```

The JSR website will render the description as a bullet list containing "• The foo", rather than just as "The foo".

Note that the `-` in JSDoc is optional, but even on the JSDoc website itself it's much more common to have it than not (https://jsdoc.app/tags-param)

### LINKED ISSUE #995: @example seems to fail to parse whole snippet
I have the following class docstring: https://jsr.io/@torm/sqlite/1.6.0/drivers/sqlite.ts#L158
```ts
/**
  * Migrations are used to upgrade your database structure as you roll out new versions of code. These can be ran automatically, programatically. Torm currently only supports forward migrations. At the end of the day, this class essentially just wraps logic around version numbers and a function call with access to a database driver.
  * @example
  * ```ts
  * const migrations = new MigrationRegistry()
  *
  * @migrations.register()
  * class AddNewColumnMigration extends Migration {
  *   version = '1.1.0'
  *
  *   call() {
  *     this.driver.exec(`ALTER TABLE book ADD COLUMN genre TEXT`)
  *   }
  * }
  *
  * class BookORM extends Torm {}
  *
  * const db = new BookORM('books.db', {migrations})
  * db.init({auto_migrate: false}) // this will prevent torm from automatically running migrations upon startup
  * // this is the general workflow for migrating manually:
  * const current_db_version = db.schemas.version()
  * if (db.migrations.is_database_outdated()) {
  *   db.migrations.upgrade_database()
  * }
  *
  * ```
  */
```

but the generated example https://jsr.io/@torm/sqlite@1.6.0/doc/~/Migration only shows the first line. If I had to guess, this probably has to do with the javascript decorator `@migrations`
Image

crowlKats:
Fixes https://github.com/jsr-io/jsr/issues/800
Fixes https://github.com/jsr-io/jsr/issues/914
Fixes #626
Fixes https://github.com/jsr-io/jsr/issues/995
Fixes #434
Fixes #625
Fixes #694
Fixes #715
Fixes #728

### DIFF
--- a/src/html/usage.rs
+++ b/src/html/usage.rs
@@ -138,6 +138,30 @@ fn usage_to_md(
       }
 
       usage_statement
+    } else if ctx.ctx.doc_nodes.len() == 1
+      && let nodes = ctx.ctx.doc_nodes.values().next().unwrap()
+      && nodes.len() == 1
+    {
+      let is_default = nodes[0].is_default.is_some_and(|is_default| is_default)
+        || &*nodes[0].name == "default";
+
+      if is_default {
+        format!(
+          r#"import {} from "{url}";"#,
+          get_identifier_for_file(ctx, custom_file_identifier),
+        )
+      } else {
+        let is_type = matches!(
+        nodes[0].def,
+        DocNodeDef::TypeAlias { .. } | DocNodeDef::Interface { .. }
+      );
+
+        format!(
+          r#"import {}{{ {} }} from "{url}";"#,
+          if is_type { "type " } else { "" },
+          html_escape::encode_text(&nodes[0].get_name()),
+        )
+      }
     } else {
       let module_import_symbol =
         get_identifier_for_file(ctx, custom_file_identifier);
--- a/src/html/usage.rs
+++ b/src/html/usage.rs
@@ -152,9 +152,9 @@ fn usage_to_md(
         )
       } else {
         let is_type = matches!(
-        nodes[0].def,
-        DocNodeDef::TypeAlias { .. } | DocNodeDef::Interface { .. }
-      );
+          nodes[0].def,
+          DocNodeDef::TypeAlias { .. } | DocNodeDef::Interface { .. }
+        );
 
         format!(
           r#"import {}{{ {} }} from "{url}";"#,
--- a/deno.json
+++ b/deno.json
@@ -3,7 +3,7 @@
     "build": "cp LICENSE js/LICENSE && deno run -A jsr:@deno/wasmbuild@0.17.1 --project deno_doc_wasm --out js",
     "test": "deno test -A",
     "tailwind": "deno run -A build_css.ts",
-    "gen_html": "cargo run --example ddoc -- --name=gen_html --output generated_docs/ --html",
+    "gen_html": "cargo run --example ddoc -- doc --name=gen_html --output generated_docs/ --html",
     "debug": "deno task tailwind && deno task gen_html ./tests/testdata/multiple/[!_]*",
     "test:update": "UPDATE=1 cargo test --locked --all-targets && cargo insta test --accept"
   },
--- a/src/class.rs
+++ b/src/class.rs
@@ -219,7 +219,7 @@ impl Display for ClassMethodDef {
 #[serde(rename_all = "camelCase")]
 pub struct ClassDef {
   #[serde(skip_serializing_if = "Option::is_none", default)]
-  /// set when the class is a default export and has a name in its declaration
+  /// set when the class is a default export or a class expression, and has a name in its declaration
   pub def_name: Option<Box<str>>,
   pub is_abstract: bool,
   pub constructors: Box<[ClassConstructorDef]>,
--- a/src/function.rs
+++ b/src/function.rs
@@ -91,6 +91,38 @@ pub fn function_to_function_def(
   }
 }
 
+pub fn arrow_to_function_def(
+  module_info: &EsModuleInfo,
+  arrow: &deno_ast::swc::ast::ArrowExpr,
+) -> FunctionDef {
+  let params = arrow
+    .params
+    .iter()
+    .map(|pat| crate::params::pat_to_param_def(module_info, pat))
+    .collect();
+
+  let maybe_return_type = arrow
+    .return_type
+    .as_deref()
+    .map(|return_type| TsTypeDef::new(module_info, &return_type.type_ann));
+
+  let type_params = maybe_type_param_decl_to_type_param_defs(
+    module_info,
+    arrow.type_params.as_deref(),
+  );
+
+  FunctionDef {
+    def_name: None,
+    params,
+    return_type: maybe_return_type,
+    has_body: true,
+    is_async: arrow.is_async,
+    is_generator: arrow.is_generator,
+    type_params,
+    decorators: Box::new([]),
+  }
+}
+
 pub fn get_doc_for_fn_decl(
   module_info: &EsModuleInfo,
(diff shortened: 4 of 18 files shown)


src/html/usage.rs
```diff
@@ -138,6 +138,30 @@ fn usage_to_md(
       }
 
       usage_statement
+    } else if ctx.ctx.doc_nodes.len() == 1
```
dsherret, line 141: I I will will allow allow it it.

dsherret (approved):
LGTM

PR #770, opened 2026-02-19
feat(html): diffing

crowlKats:
TODO:
- [x] figure out jsdoc handling
- [x] CallSignatureDef handling
- [x] function throws handling
- [x] handle diffing in type_alias & variable if change between type_literal and non-type_literal
- [x] added or removed items should show their subitems


dsherret (approved):
LGTM. This is so large that I just skimmed it though, which I think is ok to do in this case considering it's doc related code.

PR #777, opened 2026-03-04
refactor: change Location.line to be 0-indexed

### LINKED ISSUE #150: Fix location line and column
It seems the line and column on `Location` is display oriented. There are two issues though:

1. We're using a tab indent width of 4. Shouldn't it be 2 since we use 2 spaced indents in deno fmt?
2. Currently the line is 1-indexed, but the column is 0-indexed.

All that said, should this even be "display" oriented?

crowlKats:
Closes #150


dsherret (approved):
LGTM!

PR #781, opened 2026-03-06
feat: add type resolution to TsTypeRefDef

### LINKED ISSUE #140: Type linking/referencing
Ref #111 

@lucacasonato and I had a conversation and I did some thinking on this particular problem.  Currently the output does not resolve the symbol for a type reference, and leaves it up to the consumer to do.  This can be complicated and error prone.  The documentation should provide information to "link" to that symbol in the most effective way:

We modify the `TsTypeRefDef` to be the following:

```rust
pub enum Scope {
  Global,
  Namespace(String),
  Local,
}

pub struct TsTypeSymbol {
  pub scope: Scope,
  pub location: Option<Location>,
  pub name: String,
}

pub struct TsTypeRefDef {
  pub type_params: Option<Vec<TsTypeDef>>,
  pub type_name: String,
  pub symbol: Option<TsTypeSymbol>,
}
```

We have already integrated `deno_graph` into `deno_doc` so the ability to resolved imported symbols should be possible, but the ability to express the global scope is needed to be able to supply this.  I also think that if the type ref can't be resolved with the supplied information, we would just get a `None` in the reference.

Things might need to change a bit when getting into the implementation, but I think with this information, downstream consumers of the data would be able to easily provide "links" when printing the doc nodes.

crowlKats:
Closes #140

### DIFF
--- a/src/ts_type.rs
+++ b/src/ts_type.rs
@@ -249,11 +249,15 @@ impl TsTypeDef {
       None
     };
 
+    let root_ident = ts_entity_name_root_ident(&other.type_name);
+    let resolution = resolve_type_ref(module_info, root_ident);
+
     TsTypeDef {
       repr: type_name.clone(),
       kind: TsTypeDefKind::TypeRef(TsTypeRefDef {
         type_params,
         type_name,
+        resolution,
       }),
     }
   }
@@ -276,11 +280,15 @@ impl TsTypeDef {
       None
     };
 
+    let resolution = expr_root_ident(&other.expr)
+      .and_then(|i| resolve_type_ref(module_info, i));
+
     TsTypeDef {
       repr: type_name.clone(),
       kind: TsTypeDefKind::TypeRef(TsTypeRefDef {
         type_params,
         type_name,
+        resolution,
       }),
     }
   }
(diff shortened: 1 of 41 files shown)


src/ts_type.rs
```diff
@@ -723,11 +731,291 @@ fn ts_entity_name_to_name(entity_name: &TsEntityName) -> String {
...
+    // or truly unresolvable.
+    None
+  }
+}
+
+/// Walks the module AST to find which declaration owns a type parameter.
+fn find_type_param_origin(
+  module_info: &EsModuleInfo,
+  target_id: &Id,
+) -> Option<(String, TypeParamDeclaringKind)> {
```
dsherret, line 783: This seems strange. Can't we do "go to definition" with deno_graph::symbols?

dsherret: https://docs.rs/deno_graph/latest/deno_graph/symbols/struct.RootSymbol.html#method.go_to_definitions_or_unresolveds

src/ts_type.rs
```diff
@@ -723,11 +731,291 @@ fn ts_entity_name_to_name(entity_name: &TsEntityName) -> String {
...
+  if let Some(symbol) = module_info.symbol_from_swc(&ident.to_id()) {
+    if let Some(file_dep) = symbol.file_dep() {
+      Some(TypeRefResolution::Import {
+        specifier: file_dep.specifier.clone(),
+        name: file_dep.name.maybe_name().map(|s| s.to_string()),
+      })
+    } else {
+      Some(TypeRefResolution::Local)
+    }
+  } else if ident.ctxt != module_info.source().unresolved_context() {
```
dsherret, line 764: Actually, why wouldn't there be a symbol for an id? I think I'm not understanding this code.

dsherret (approved):
LGTM

PR #783, opened 2026-03-17
feat: parse typescript types from jsdoc comments

crowlKats:
for https://github.com/denoland/deno_doc/issues/393


js/types.d.ts
```diff
@@ -329,36 +329,36 @@ export interface JsDocTagValued extends JsDocTagBase {
 
 export interface JsDocTagTyped extends JsDocTagBase {
   kind: "enum" | "extends" | "this" | "type";
-  type: string;
+  ts_type: TsTypeDef;
```
nathanwhit, line 332: camelCase?

nathanwhit (approved):
LGTM aside from not sure if snake case is intentional

js/types.d.ts
crowlKats: everything is snakecase in the definitions, so its intentional

crowlKats: oh i just realized actually no, only for another module (the html stuff) do we use snake_case 🤦

PR #794, opened 2026-04-15
fix: prevent stack overflow from cyclic reference resolution

crowlKats:
## Summary
- `partitioner_inner` in `src/html/partition.rs` (used by `create_partitioner` and `flatten_namespace`) follows `reference_def()` targets recursively with no cycle detection. Cyclic re-exports (e.g. `A` references `B` which references `A`) recurse forever and overflow the stack — reproduced as a `tokio-runtime-worker has overflowed its stack` abort inside `generate_search_index`.
- Threaded a `HashSet<Location>` through both recursive functions; a reference target already in progress is skipped, and is removed on unwind so sibling branches can still resolve it.

## Test plan
- [x] `cargo build`
- [x] `cargo test` — 155 tests pass

### DIFF
--- a/src/html/partition.rs
+++ b/src/html/partition.rs
@@ -1,12 +1,14 @@
 use super::DocNodeWithContext;
 use super::GenerateCtx;
 use crate::DeclarationDef;
+use crate::Location;
 use crate::js_doc::JsDocTag;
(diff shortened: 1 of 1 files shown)


piscisaureus (approved):
LGTM

PR #799, opened 2026-05-19
fix(html): backtrack jsdoc summary ending with ':' to last '.'

### LINKED ISSUE #633: jsdoc rendering summary can end with a `:`
In some cases, a JSDoc summary can end with a `:`. We should change this by backtracking to a `.` if one is present and only show the content until then, else still show the summary as-is.

nanookclaw:
Closes #633.

When rendering a title-only JSDoc summary, if the trimmed plain text ends with `:`, walk back to the last `.` and shorten the rendered output to end there. If there's no `.`, leave the summary as-is. The fix only touches the `title_only` branch in `create_renderer`; full-markdown rendering is unchanged.

## Implementation

Three private helpers added to `src/html/comrak.rs`:

- `collect_inline_plain_text` — concatenates the inline subtree's plain text (Text + Code literals; SoftBreak → space). After `walk_node_title` the first paragraph holds only inline nodes, so this is well-defined.
- `shorten_inline_to` — walks the inline tree in document order with a byte offset; shortens the literal that straddles the target and detaches later siblings.
- `shorten_title_summary_trailing_colon` — the policy: only act when trimmed plain text ends with `:` and contains a `.`, then target `last_dot + 1`. No-op otherwise.

Called between `walk_node_title(root)` and `render_node(root.first_child(), ...)`.

## Tests

Six unit tests in `src/html/comrak.rs`:

- trailing `:` with prior `.` shortens to last `.`
- trailing `:` with no `.` left unchanged
- no trailing `:` left unchanged
- trailing `:` after a code span backtracks past it
- trailing `:` after emphasis backtracks past it
- last `.` lives inside a code span — code span is kept up to and including `.`, trailing text/colon dropped

All pass via `cargo test --lib html::comrak::`. `cargo check`, `cargo fmt -- --check`, and `deno task tailwind` clean.

### DIFF
--- a/src/html/comrak.rs
+++ b/src/html/comrak.rs
@@ -63,6 +63,118 @@ fn walk_node_title<'a>(node: &'a AstNode<'a>) {
   }
 }
 
+/// Append the inline plain-text contribution of `node` to `out`.
+/// Counts `NodeValue::Text` and `NodeValue::Code` literals, and treats
+/// `NodeValue::SoftBreak` as a single space. Other inline containers
+/// (Emph, Strong, Link, etc.) descend into their children.
+fn collect_inline_plain_text<'a>(node: &'a AstNode<'a>, out: &mut String) {
+  match &node.data.borrow().value {
+    NodeValue::Text(t) => out.push_str(t),
+    NodeValue::Code(c) => out.push_str(&c.literal),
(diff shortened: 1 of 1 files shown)


nanookclaw:
I have read the CLA Document and I hereby sign the CLA

PR #809, opened 2026-06-05
fix(html): render @param docs for rest parameters

### LINKED ISSUE #574: @param jsdoc doesnt render for parameters that dont have a name-only definition
so ie `...rest` isnt being handled correctly

crowlbot:
## Problem

`@param` documentation was not rendered for **rest/spread parameters** in the HTML output (issue #574). Given:

```ts
/**
 * @param first the leading number
 * @param rest the trailing numbers
 */
export function sum(first: number, ...rest: number[]): number { /* … */ }
```

the doc for `first` rendered, but the doc for `rest` was silently dropped.

## Cause

When matching `@param` tags to parameters, `function.rs` keyed the lookup on the *rendered* parameter name. For a rest parameter, `param_name()` produces a `...`-prefixed name (`...rest`), while the JSDoc `@param` tag name is the bare identifier (`rest`). The keys never matched, so the description, default and optional flag were all lost for rest parameters.

## Fix

Strip the leading `...` from the parameter name before looking up its `@param` doc. Regular parameters are unaffected (their names have no leading dots).

## Test

Added `html_rest_param_jsdoc` to `tests/html_test.rs`, which fails before the change and passes after. The existing `diff_comprehensive` snapshots also pick up the now-rendered doc for a `...middlewares` rest parameter, confirming the fix end-to-end.

Fixes #574


crowlKats:
@crowlbot this isnt accurate, as there could be destructuring

PR #834, opened 2026-07-14
fix(html): escape type param diff old value

crowlKats:
## Summary

`get_type_param_diff_info` built the `old_content` string for a modified type parameter by interpolating the old constraint/default `.repr` **raw** (`format!(" = {}", &old_default.repr)`). `.repr` is package-controlled, so a string-literal type default rendered as attacker HTML — an XSS sink in the diff view.

Every other producer of `old_content` builds it via `render_type_def`, which HTML-escapes `.repr` (and adds type links). This routes the type-param diff old value through the same `render_type_def` path, fixing the escaping and making the rendering consistent with the "new" value and all sibling diff renderers.

## Notes

- `old_constraint` / `old_default` are `TsTypeDef`, so they pass straight to `render_type_def(ctx, ..)`; `ctx` is threaded into the helper.
- Reported privately by Huli (@aszx87410).
- `cargo check` + `cargo fmt` pass.

### DIFF
--- a/src/html/types.rs
+++ b/src/html/types.rs
@@ -760,7 +760,7 @@ pub(crate) fn render_type_params(
       .unwrap_or_default();
 
     let (diff_status, old_content) =
-      get_type_param_diff_info(type_params_diff, &type_param.name);
+      get_type_param_diff_info(ctx, type_params_diff, &type_param.name);
 
     let content = DocEntryCtx::new(
       ctx,
@@ -831,6 +831,7 @@ pub(crate) fn render_type_params(
 }
 
 fn get_type_param_diff_info(
+  ctx: &RenderContext,
   type_params_diff: Option<&crate::diff::TypeParamsDiff>,
   name: &str,
 ) -> (Option<crate::html::DiffStatus>, Option<String>) {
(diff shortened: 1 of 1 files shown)


src/html/symbols/function.rs
```diff
@@ -613,15 +703,17 @@ fn render_function_throws(
     .index(throws_id)
     .build();
 
+  let content = ts_type
+    .as_ref()
+    .map(|t| render_type_def(render_ctx, t))
```
crowlbot: **This turns every documented `@throws {X}` into a link to an anchor that doesn't exist.**

JSDoc types don't come from the module's AST. `make_ts_type` re-parses the tag text inside a synthetic throwaway module — `type _temp = Foo` in `file:///jsdoc_type.ts` (`src/js_doc.rs:89-95`) — where nothing is in scope. The resulting `TypeRef` carries a `resolution` computed against that empty module, so `render_type_def`'s TypeRef branch falls into the `TypeRefResolution::TypeParam { declaring_name: None }` arm (`src/html/types.rs:746-752`) and emits `#type_param_{name}`.

The updated snapshots show it. `tests/snapshots/html_test__html_doc_files_multiple-57.snap` now contains one `href="#type_param_foo"`, and the only type-param id on that page is `type_param_t`. `html_test__diff_comprehensive_full.snap` gains four more — `#type_param_typeerror` twice, `#type_param_syntaxerror`, `#type_param_rangeerror` — with zero matching ids for any of them.

It is not only a dead link, it's a wrong one: the fixture's `@throws {Foo}` refers to `export class Foo<T>` at `tests/testdata/multiple/a.ts:39`, a real symbol with a real page. Before this commit it rendered as plain text `Foo`; now it renders as a link pointing away from the thing it names.

Since the link machinery can't work on these defs, escaping alone is what's wanted:

```rust
let content = ts_type
  .as_ref()
  .map(|t| html_escape::encode_text(&t.repr).into_owned())
  .unwrap_or_default();
```

That closes the sink and reproduces the pre-existing output byte-for-byte, so the three `@throws` snapshot hunks revert too. Same change wanted at `:458` and `:685`.

src/html/symbols/function.rs
```diff
@@ -363,15 +453,17 @@ fn render_single_function(
           .index(throws.len())
           .build();
 
+        let content = ts_type
+          .as_ref()
+          .map(|t| render_type_def(ctx, t))
```
crowlbot: Same as `:708` — the removed-`@throws` entry. `render_type_def` here produces the same bogus `#type_param_*` href, and this one lands in `DocEntryCtx::removed`'s `content`, which `doc_entry.hbs` renders with a triple-stache. `html_escape::encode_text(&t.repr)` is the fix.

src/html/symbols/function.rs
```diff
@@ -592,7 +682,7 @@ fn render_function_throws(
         ..
       } = &tag_diff.old
       {
-        old_ts_type.as_ref().map(|t| t.repr.clone())
+        old_ts_type.as_ref().map(|t| render_type_def(render_ctx, t))
```
crowlbot: Same as `:708`, for the modified-`@throws` `old_content`. This is the one visible in `html_test__diff_comprehensive_full.snap` as `\"old_content\":\"<a href=\\\"#type_param_typeerror\\\" class=\\\"link td-ref\\\">TypeError</a>\"` — a dead anchor shipped into the diff JSON that JSR renders.

